PR #11, opened 2024-04-16
driver: keep shim alive to report exit code

### LINKED ISSUE #9: driver: protect against re-attaching to unrelated process
There's an edge case we need to be careful of. If the Nomad client restarts the exec2 driver plugin also restarts. During startup it "re-attaches" to the now-orphaned `unshare` parent process of the tasks it previously launched. This is done by keeping a reference to the PID. However it is possible the original task process dies and a new process is spawned and assigned the same PID by the kernel.

We need to avoid re-attaching to a process that is now our own. We could do this by simply comparing the process command (and arguments) after the restart, ensuring it matches the command invoked to launch the task.

shoenig:
This PR fixes a design flaw where the exit code of a task process would
not be retrievable after a Nomad client restart. This is critical for
batch jobs, the success or failure of which are determined by their exit
code.

The fix is basically to have the exec2-shim process remain alive as a
parent of the task process and record the exit code to a known file
inside the task directory. That file is then read back by the plugin
after the task process exits. This path is only necessary in the reattach
case; wait.go now splits logic between waiting on a child and waiting
on an orphan. The orphan case makes use of the recent pidfd_open syscall
to acquire an FD on which we call poll, which returns when the process
exits. Unfortunately Linux does not yet provide a way to retrieve the
exit status information via the pidfd_* set of syscalls, and so we fallback
to the shim + file hack.

While fixing this I also realized the FIFO filepaths for logging were
incorrect - we were using the host-perspective paths instead of the
task mounts directory perspective. This is a requirement for working in
a production - hardened environment, and the bug became apparent because
the tests would no longer pass without the fix. The reason is because
before, the FIFOs were opened by the plugin (running as root) and the FD's
simply passed through unshare upon launch. Now, the shim opens the FIFOs
itself and would fail to do so with the incorrect host-perspective paths
due to them being owned by root.

Closes #9

### DIFF
--- a/e2e/jobs/cgroup.hcl
+++ b/e2e/jobs/cgroup.hcl
@@ -26,7 +26,7 @@ job "cgroup" {
       config {
         command = "/usr/bin/cat"
         args    = ["/proc/self/cgroup"]
-        unveil  = ["r:/proc/self/cgroup"]
+        unveil  = ["r:/proc"]
       }
 
       resources {
--- a/pkg/resources/process/wait.go
+++ b/pkg/resources/process/wait.go
@@ -4,7 +4,7 @@
 package process
 
 import (
-	"fmt"
+	"errors"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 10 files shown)


e2e/jobs/cgroup.hcl
```diff
@@ -26,7 +26,7 @@ job "cgroup" {
       config {
         command = "/usr/bin/cat"
         args    = ["/proc/self/cgroup"]
-        unveil  = ["r:/proc/self/cgroup"]
+        unveil  = ["r:/proc"]
```
shoenig, line 29: `/proc/self` is the shim process; now that we fork+exec instead of just exec, `cat` is a child and cannot see _it's_ `/proc/self`, which will not have been made available by landlock. There's Tylenol in the kitchen.

pkg/resources/process/wait.go
```diff
@@ -69,47 +72,101 @@ func (w *procWaiter) wait(ch chan<- *drivers.ExitResult) {
...
+		Fd:     pidfd,
+		Events: unix.POLLIN,
+	}}
+	const timeout = -1 // infinite
+
+	// no need to check for an error
+	_, _ = unix.Poll(pollFD, timeout)
+
+	// retrieve code from file
+	source := filepath.Join(w.taskdir, "local", exitStatusFile)
```
schmichael, line 135: Don't you need to check for this *before* using the pidfd in case a *new* process has started with the same pid as the original task process?

pkg/shim/z_shim_cmd.go
```diff
@@ -4,80 +4,115 @@
...
+		stdout, stderr, err := util.OpenPipes(outPipePath, errPipePath)
+		if err != nil {
+			subproc.Print("failed to open output pipes: %v", err)
+			return ExitBadLogging
+		}
+
+		// give ourselves a way to write to the stderr pipe for printing fatal errors
+		debug := func(format string, args ...any) {
+			_, _ = io.WriteString(stderr, fmt.Sprintf(format, args...))
+		}
```
tgross, line 80: Just for my clarity, this it the same named pipe we eventually give to the workload as its stderr? So these logs will show up in `nomad alloc logs`?

pkg/resources/process/wait.go
tgross: Yeah, I think we need to:
* grab the pidfd
* check the exit file is empty while we're holding the pidfd
* then poll the pidfd
* then read the file value when we're done

Also, just double-checking but this PID is the PID of the shim process, right? So we're guaranteed that if the pidfd unblocks that the exit status file has already been written and there's no race between reading and writing?

pkg/resources/process/wait.go
```diff
@@ -69,47 +72,101 @@ func (w *procWaiter) wait(ch chan<- *drivers.ExitResult) {
...
-		}
+		ch <- unrecoverable("task process handle is gone", err)
+		return
+	}
+
+	pidfd, err := openProcessFD(w.pid)
+	if err != nil {
+		ch <- unrecoverable("task process file descriptor cannot be opened", err)
+		return
+	}
```
tgross, line 123: If the task exits quickly (ex the shim process fails during startup), it'll have written any to the named pipe logs but we won't have the exit code? I feel like what I'm asking is supposed to be covered given your description:

> wait.go now splits logic between waiting on a child and waiting on an orphan.

But I'm not sure I understand where that split is happening here.

pkg/resources/process/wait.go
shoenig: Great ideas! I added 067f9a5 which rewrites this section to take the already-dead (and potentially reborn) process PID into account. I believe doing so also elegantly resolves https://github.com/hashicorp/nomad-driver-exec2/issues/9.

This PID is of the `unshare` process which is a parent of the `shim`. Since `unshare` is configured to `SIGKILL` its children upon its own death the shim will also be dead (and it shouldn't die until the shim exits anyway). I think in practice, we're fine. 

In the next hackathon I'd like to sit down and write a tool that combines `unshare`, `nsenter`, and the landlock parts of the shim into one tool. Probably in Rust so there's no Go runtime overhead.

pkg/shim/z_shim_cmd.go
tgross: I recently was chatting with @schmichael (IIRC) about trying to pull the `executor.log` for the exec v1 driver into the alloc logs too! So cool that we're already doing it here, and I think this reduces my worry that we could be exposing something sensitive in the logs this way.

tgross (approved):
LGTM!

PR #12, opened 2024-04-17
build: initial steps for building via CRT

shoenig:


.release/security-scan.hcl
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+binary {
+  secrets    = true
```
tgross, line 5: Is this newer syntax? The `nomad` repository has a `binary.secrets` block: https://github.com/hashicorp/nomad/blob/main/.release/security-scan.hcl#L18-L20

tgross (approved):
LGTM! :shipit:

.release/security-scan.hcl
shoenig: no idea about new-ness, but it's from the reference file

https://github.com/hashicorp/crt-core-helloworld/blob/main/.release/security-scan.hcl

PR #14, opened 2024-04-18
e2e: add a Java example to the e2e test suite

shoenig:
This PR adds an example of compiling and running a Java program using the exec2 task driver to the e2e suite. Depending on your version of Java you'll need to `unveil` some path under `/etc/` for reading - for OpenJDK that will be something like `/etc/java-17-openjdk`.


tgross (approved):
LGTM! 

Fun little trick to have the hello world compiled right in the test :grin:

PR #21, opened 2024-05-14
e2e: temporarily use nobody user for resources test

shoenig:
Set the user=nobody while figuring out a logging permission fix. We
want this so that we can merge https://github.com/hashicorp/nomad-driver-exec2/pull/16
which is almost certainly going to be conflated.


tgross (approved):
LGTM :+1:

PR #23, opened 2024-05-16
e2e: add a test to ensure we can run envoy

shoenig:
A "hello world" envoy job - demonstrating the use of `unveil` to allow access to `/dev/shm` which envoy requires.

I thought about doing a full `countdash` example but that requires running consul and whatnot. Maybe we can add exec2 examples to Nomad's e2e suite and do it there.


tgross (approved):
LGTM!

Agreed that the full Connect example is probably best done in the Nomad nightly E2E, as that already has all the WI configuration setup done.

Issue #15, opened 2024-04-18
driver: secrets directory is broken


shoenig:
While trying to launch envoy as a Connect sidecar, the path is relative to the host allocation directory tree, rather than to the mounts directory tree.

```
[2024-04-18 19:19:19.638][11][critical][main] [source/server/server.cc:414] error initializing config '  /tmp/NomadClient1208852000/f869a60a-6e84-bccc-46e1-780f5b2371ea/connect-proxy-count-api/secrets/secrets/envoy_bootstrap.json': Invalid path: /tmp/NomadClient1208852000/f869a60a-6e84-bccc-46e1-780f5b2371ea/connect-proxy-count-api/secrets/secrets/envoy_bootstrap.json
[2024-04-18 19:19:19.638][11][info][main] [source/server/server.cc:1029] exiting
```

Taking a dump of the environment, looks like we may just be not setting the NOMAD_SECRETS_DIR correctly:

```
➜ nomad alloc logs ed
NOMAD_MEMORY_LIMIT=16
NOMAD_CPU_LIMIT=100
NOMAD_SHORT_ALLOC_ID=ed5b4909
NOMAD_ALLOC_ID=ed5b4909-7545-31ae-a635-8a87fbbb00f1
HOME=/nonexistent
NOMAD_LICENSE_PATH=/etc/hashicorp.d/nomad.license
NOMAD_TASK_DIR=/tmp/NomadClient1208852000/ed5b4909-7545-31ae-a635-8a87fbbb00f1-task/task/local
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/shoenig/.bin
NOMAD_NAMESPACE=default
TMPDIR=/tmp/NomadClient1208852000/ed5b4909-7545-31ae-a635-8a87fbbb00f1-task/task/tmp
NOMAD_ALLOC_INDEX=0
NOMAD_ALLOC_DIR=/tmp/NomadClient1208852000/ed5b4909-7545-31ae-a635-8a87fbbb00f1-task/alloc/alloc
NOMAD_JOB_ID=env
NOMAD_GROUP_NAME=group
NOMAD_JOB_NAME=env
NOMAD_CPU_CORES=
NOMAD_REGION=global
NOMAD_SECRETS_DIR=/tmp/NomadClient1208852000/ed5b4909-7545-31ae-a635-8a87fbbb00f1/task/secrets/secrets
[...]

PR #24, opened 2024-05-16
e2e: add a test demonstrating secrets/identity work

shoenig:
Closes https://github.com/hashicorp/nomad-driver-exec2/issues/15 

(which was fixed in https://github.com/hashicorp/nomad/pull/20589)


tgross (approved):
LGTM!

Issue #4, opened 2024-04-09
release: use CRT to publish releases 


shoenig:

david-yu:
It would be nice if we could assume the default `/opt/nomad/data` dir for installs on nomad for plugin_dir see https://github.com/hashicorp/nomad/blob/main/.release/linux/package/etc/nomad.d/nomad.hcl#L6. 

Given the assumption, it would be great to also be able to install via yum or apt packaging so one could easily install `nomad` and `exec2` using something like

```
$ apt-get install nomad nomad-driver-exec2

```

PR #25, opened 2024-05-20
build: create linux package during CRT build

shoenig:
This PR creates .deb and .rpm Linux packages of the nomad-driver-exec2
task driver. Given limitations of our packaging system the driver gets
installed in /usr/bin/ but then gets linked to /opt/nomad/plugins/ during
the install steps (and likewise removed during uninstall). This way the
plugin Just Works with the default configuration of Nomad.

Closes https://github.com/hashicorp/nomad-driver-exec2/issues/4


schmichael (approved):

david-yu:
Awesome, thank you!

PR #32, opened 2024-06-04
config: unveil NOMAD_* paths if unveil_defaults is set

shoenig:
This PR changes the driver such that NOMAD_TASK_DIR, NOMAD_ALLOC_DIR, and
NOMAD_SECRETS_DIR are automatically unveil'd with rwxc permissions if
the plugin `config.unveil_defaults` value is set. Previously we would do
this if and only if the task's `config.unveil` value was empty, which does
not make sense.

Also re-enable test cases that were disabled when we fixed the /tmp
unveil behavior. The unit tests had to be tweaked to actually resolve
the environment variables at run time because the testing harness does
not do that and figuring out how to make it work was too hard.


plugin/driver_test.go
```diff
@@ -320,33 +320,34 @@ func TestFunctional_cases(t *testing.T) {
...
+		},
+		{
+			name:           "write to secrets directory",
+			user:           "nomad-80000",
+			command:        "sh",
+			unveilDefaults: true,
+			unveilPaths:    []string{"r:/etc/hosts"},
+			args:           []string{"-c", "cp /etc/hosts ${NOMAD_SECRETS_DIR}"},
+			exp:            &drivers.ExitResult{ExitCode: 0},
+		},
```
tgross, line 350: It probably seems "obvious" now, but if we had a test that exercised the reverse case (that `NOMAD_*_DIR` is not unveiled when `unveil_defaults=false`), that test would illustrate that it's by design and not by accident.

tgross (approved):
LGTM

schmichael (approved):

PR #33, opened 2024-06-05
driver: add tests that fail to write into the tasks own sandbox

shoenig:
If unveil_defaults is not set, it should not be possible to write into
the task's own sandbox. In this case the task should be configured with
its own sandbox as one of its unveil paths.


tgross (approved):
LGTM!

PR #35, opened 2024-06-05
readme: clarify what bin files are auto unveiled

### LINKED ISSUE #19: docs are ambiguous on whether command is automatically unveiled
The examples given in the README and examples seem to suggest that the command is unveiled automatically, but looking at the sandbox code (ref [`sandbox.go#L51-L55`](https://github.com/hashicorp/nomad-driver-exec2/blob/v0.1.0-alpha.1/pkg/shim/sandbox.go#L51-L55)) it looks like the sandbox automatically unveils `/bin`, `/usr/bin`, and `/usr/local/bin`. We should clarify this behavior in the README.

shoenig:
Fixes #19

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ dependent, and can be disabled or customized in plugin config.
 The default set of default paths are listed below. These paths are enabled only
 if they are found to exist at the time of the task launching.
 
+##### bin files
+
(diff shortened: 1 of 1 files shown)


tgross (approved):
LGTM!

PR #38, opened 2024-06-06
plugin: able to use task sandbox tmp directory

### LINKED ISSUE #37: using TMPDIR results in permission denied problems
Nomad creates a `tmp` directory for each task sandbox. This directory gets mounted automatically for tasks making use of unveil filesystem isolation, and we intend for tasks to make use of it by referencing the standard TMPDIR environment variable. However, the structure of the real task sandbox is such that the tmp directory is not part of the _task_ directory or _alloc_ directory (or _secrets_ directory), and is therefore currently not actually accessible by tasks by default.

```
[drwxr-xr-x root    ]  .
├── [drwx--x--x root    ]  alloc_mounts
│   └── [drwx--x--- 88838   ]  e0fc21ec-ffd3-5f4d-b996-f27af87fb664-mytask
│       ├── [drwxrwxrwx nobody  ]  alloc                     <- RWXC by default
│       │   ├── [drwxrwxrwx nobody  ]  data
│       │   ├── [drwxrwxrwx nobody  ]  logs
│       │   │   ├── [-rw-r--r-- root    ]  mytask.stderr.0
│       │   │   └── [-rw-r--r-- root    ]  mytask.stdout.0
│       │   └── [drwxrwxrwx nobody  ]  tmp
│       ├── [drwxrwxrwx nobody  ]  local                    <- RWXC by default
│       ├── [drwxr-xr-x root    ]  private
│       ├── [drwx--x--- 88838   ]  secrets                  <- RWXC by default
│       └── [drwxrwxrwt nobody  ]  tmp                      <- oops
```

We can see the problem by running a simple `mktemp` command.

```hcl
job "tmp" {
  type = "batch"

  group "group" {

    task "mktemp" {
      driver = "exec2"

      config {
        command = "mktemp"
      }

      resources {
        cpu    = 300
        memory = 32
      }
    }

    restart {
      attempts = 0
      mode     = "fail"
    }
  }
}
```

Output

```
➜ nomad alloc logs -stderr 57
/usr/bin/mktemp: failed to create file via template ‘/tmp/NomadClient240506396/57197d96-b707-d797-fc63-462efbc8cff0-mktemp/tmp/tmp.XXXXXXXXXX’: Permission denied
```

shoenig:
This PR fixes a bug where the tmp directory created on behalf of a task
in its sandbox directory was not accessible to tasks. Adds some tests
to make sure this keeps working.

Fixes #37

### DIFF
--- a/plugin/driver.go
+++ b/plugin/driver.go
@@ -515,6 +515,8 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
 		unveil = append(unveil, "rwxc:"+driverTaskConfig.Env["NOMAD_TASK_DIR"])
 		unveil = append(unveil, "rwxc:"+driverTaskConfig.Env["NOMAD_ALLOC_DIR"])
 		unveil = append(unveil, "rwxc:"+driverTaskConfig.Env["NOMAD_SECRETS_DIR"])
+		parent := filepath.Dir(driverTaskConfig.Env["NOMAD_TASK_DIR"])
+		unveil = append(unveil, "rwxc:"+filepath.Join(parent, "tmp"))
 	}
 
 	if len(taskConfig.Unveil) > 0 {
--- a/e2e/basic_test.go
+++ b/e2e/basic_test.go
@@ -163,6 +163,20 @@ func TestBasic_Env(t *testing.T) {
 	must.RegexMatch(t, containsUserRe, output)
 }
 
+func TestBasic_Mktemp(t *testing.T) {
+	ctx := setup(t)
+	defer purge(t, ctx, "mktemp")()
+
+	_ = run(t, ctx, "nomad", "job", "run", "./jobs/mktemp.hcl")
+	statusOutput := run(t, ctx, "nomad", "job", "status", "mktemp")
+	alloc := allocFromJobStatus(t, statusOutput)
+
+	// contains reasonable mktemp output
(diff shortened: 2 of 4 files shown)


plugin/driver.go
```diff
@@ -515,6 +515,8 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
 		unveil = append(unveil, "rwxc:"+driverTaskConfig.Env["NOMAD_TASK_DIR"])
 		unveil = append(unveil, "rwxc:"+driverTaskConfig.Env["NOMAD_ALLOC_DIR"])
 		unveil = append(unveil, "rwxc:"+driverTaskConfig.Env["NOMAD_SECRETS_DIR"])
+		parent := filepath.Dir(driverTaskConfig.Env["NOMAD_TASK_DIR"])
+		unveil = append(unveil, "rwxc:"+filepath.Join(parent, "tmp"))
```
tgross, line 519: This is probably ignorance on my part, but this surprises me; if the `tmp` dir is a child of the `NOMAD_TASK_DIR`, why isn't it automatically accessible by our unveiling the `NOMAD_TASKD_DIR`?

tgross: :facepalm: nevermind... `parent` is the _parent_ of the `NOMAD_TASK_DIR`.

tgross (approved):
LGTM!

plugin/driver.go
shoenig: yeah that caught me by surprise too, but I don't make the rules

PR #50, opened 2024-08-22
add newlines to stderr messages and fix error logs

schmichael:
Bandwidth can't return an error, so that code is unreachable.

`Failed to parse options` is fairly trivial to hit though by attempting to use `unveil` in a task without having `unveil_by_task = false` in your agent config.

Nomad does not truncate stderr logs between task restarts, so I added a newline to debug statements from the shim command. Without newlines everything gets written to one long line on restart.


jrasell (approved):
LGTM.

shoenig (approved):

Issue #54, opened 2024-09-17
add /usr/lib64 to default set of unveil paths


shoenig:
https://github.com/shoenig/go-landlock/issues/83

`/usr/lib64` is not part of the default set of unveil paths for shared objects, but it should be (especially important on the SLES side)

PR #55, opened 2024-09-17
deps: update nomad and go-landlock and test dependencies

shoenig:
Incorporates fix from https://github.com/shoenig/go-landlock/pull/86

Fixes: #54 

Note: can't reference a tagged version of Nomad until 1.9 drops, which contains a commit not being backported that fixes things for running exec2 tests. (If I could remember which commit is the fix, we _could_ backport it...)


jrasell (approved):
LGTM!

PR #60, opened 2024-10-21
plugin: make use of Topology object from driver fingerprint

shoenig:
This PR fixes exec2 to make use of the Topology object created during
driver fingerprinting for computation of CPU resources. This fixes cases
where the naive detection being used before did not work, for example
on ARM CPUs.

### DIFF
--- a/pkg/resources/utilization.go
+++ b/pkg/resources/utilization.go
@@ -4,12 +4,10 @@
 package resources
 
 import (
-	"os"
-	"regexp"
-	"strconv"
-	"strings"
 	"sync"
 	"time"
+
+	"github.com/hashicorp/nomad/client/lib/cpustats"
 )
 
 type MicroSecond uint64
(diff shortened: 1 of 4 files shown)


shoenig:
Spot checking on EC2 arm instance

```
ubuntu@ip-172-31-26-226:~$ lscpu | head -n10
Architecture:                         aarch64
CPU op-mode(s):                       32-bit, 64-bit
Byte Order:                           Little Endian
CPU(s):                               4
On-line CPU(s) list:                  0-3
Vendor ID:                            ARM
Model name:                           Neoverse-N1
Model:                                1
Thread(s) per core:                   1
Core(s) per socket:                   4
```

task launches just fine 

```
ubuntu@ip-172-31-26-226:~$ nomad alloc status 5f | grep Health
Deployment Health   = healthy
```

example job

```hcl
job "bug" {

  group "group" {

    task "sleep" {
      driver = "exec2"

      config {
        command = "sleep"
        args = ["infinity"]
      }

      resources {
        cores = 2
        memory =128
      }
    }
  }
}
```

schmichael (approved):

tgross (approved):
LGTM :+1:

PR #67, opened 2025-04-10
update Nomad to 1.10.0 GA and Go 1.24.2

tgross:
This changeset also pins libcap to 1.2.71. See https://github.com/shoenig/go-landlock/issues/91 for details.


tgross:
For the failing tests, I used this modified version of execsnoop to track down the failure to the exec-shim process, and from there I was able to inject bogus error codes to stumble my way thru until I found the error was coming from the landlock library. And from there I found https://github.com/shoenig/go-landlock/issues/91 and noticed that we'd accidentally bumped the library version of libcap. Pinning that should fix the driver.

<details><summary>execsnoop.by</summary>

```awk
#!/usr/bin/bpftrace
/*
 * execsnoop.bt   Trace new processes via exec() syscalls.
 *                For Linux, uses bpftrace and eBPF.
 *
 * This traces when processes call exec(). It is handy for identifying new
 * processes created via the usual fork()->exec() sequence. Note that the
 * return value is not currently traced, so the exec() may have failed.
 */

#ifndef BPFTRACE_HAVE_BTF
#include <linux/sched.h>
#endif

BEGIN
{
        printf("%-15s %-7s %-7s %s\n", "TIME", "PID", "PPID", "ARGS");
}

tracepoint:syscalls:sys_enter_exec*
{
        $task = (struct task_struct *)curtask;
        printf("%15s %-7d %-7d ", strftime("%H:%M:%S.%f", nsecs), pid, $task->real_parent->pid);
        join(args.argv);
}

tracepoint:sched:sched_process_exit {
    $task = (struct task_struct *)curtask;
    if ($task->pid == $task->tgid) {
        printf("- %15s %-7d exit(%d) %s\n", strftime("%H:%M:%S.%f", nsecs), pid, $task->exit_code >> 8, comm);
    }
}
```

</details>

PR #69, opened 2025-06-13
Conditional PWR signal for macos

DeanHnter:
The current PWR signal inclusion causes build errors as its not available on darwin builds.

`# github.com/hashicorp/nomad-driver-exec2/pkg/resources/process ../../../go/pkg/mod/github.com/hashicorp/nomad-driver-exec2@v0.1.0/pkg/resources/process/signal.go:43:18: undefined: syscall.SIGPWR (base)  % go build`

### DIFF
--- a/pkg/resources/process/signal.go
+++ b/pkg/resources/process/signal.go
@@ -39,8 +39,10 @@ func parse(s string) syscall.Signal {
 		return syscall.SIGTERM
 	case "sigstop":
 		return syscall.SIGSTOP
-	case "sigpwr":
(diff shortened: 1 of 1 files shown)


tgross (changes requested):
@DeanHnter this driver relies on Linux-only features (Landlock) and is only supported on Linux. This shouldn't a problem when cross-compiling because you'd use the correct tags there. Can you give a little more context on what you're trying to do here?

tgross:
Ok, but thanks for the attempt anyways! :grin:

PR #86, opened 2026-05-07
update copyright headers

tgross:
It doesn't particularly make sense that headers would be updated if we're not updating the contents of a given file, but this is causing failing tests in a private fork.

- [ ] If a change needs to be reverted, we will roll out an update to the code within 7 days.

### DIFF
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -1,4 +1,4 @@
-# Copyright IBM Corp. 2024, 2025
+# Copyright IBM Corp. 2024, 2026
 # SPDX-License-Identifier: MPL-2.0
 
 schema = "1"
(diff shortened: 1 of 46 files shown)


.release/ci.hcl
```diff
@@ -1,4 +1,4 @@
-# Copyright IBM Corp. 2024, 2025
+# Copyright IBM Corp. 2024, 2026
```
gulducat, line 1: I don't know a lot about IBM's requirements around this, but this reads as "2024 and 2026" rather than "2024 through 2026"... intuitively, seems like it should be

```
# Copyright IBM Corp. 2024-2026
```

?

gulducat (approved):

.release/ci.hcl
tgross: I don't make the rules, this is how it's being done everywhere. :grin:

PR #89, opened 2026-05-12
use `os.Root` when opening or modifying logging FIFOs

tgross:
If an `exec2` task restarts after another task has swapped its FIFO out for a symlink to a sensitive file, the `exec2` driver will read that contents of that file out into the logs. It will also change the ownership of the file so that it's owned by the dynamic uid/gid of the task.

Use `os.Root` to verify that the FIFO is not traversing out of the allocation log directory, and make the logs directory read-only for the task.

Ref: https://hashicorp.atlassian.net/browse/SECVULN-42948
Ref: https://github.com/hashicorp/nomad-driver-exec2-private/pull/1

- [ ] If a change needs to be reverted, we will roll out an update to the code within 7 days.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+SECURITY:
+* Protect logging FIFO from symlink swap attacks (CVE-2026-8052) [GH-87](https://github.com/hashicorp/nomad-driver-exec2/issues/87)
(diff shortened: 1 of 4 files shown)


pkazmierczak (approved):
LGTM 👍

gulducat (approved):
LGTM!

PR #96, opened 2026-08-03
support ambient capabilities

### LINKED ISSUE #79: Support allow_caps
Right now, you can't run a binary that listen to a port under 1024 because the binary is ran with a non root user (which is expected).

Allow the ability to specify capabilities like the exec plugin: https://developer.hashicorp.com/nomad/docs/deploy/task-driver/exec#allow_caps

ritesh-harihar:
Fixes: https://github.com/hashicorp/nomad-driver-exec2/issues/79

_**Summary**_

Adds two new configuration fields  `allow_caps ` and `alloc_caps` that let operators and job authors selectively grant Linux ambient capabilities to exec2 tasks. This is the capability model used by Nomad's built-in exec driver, now brought to exec2.

The primary use case is letting a non-root dynamic workload user bind a privileged port (e.g. port 80) via CAP_NET_BIND_SERVICE without running as root — something that was structurally impossible before this change.

_**Before This Change**_

**The problem**
exec2 ran every task through a three-level process chain, with unshare --setuid responsible for the uid/gid transition:

```
Nomad agent (root)
  └─ unshare --setuid=80000 --setgid=80000 --ipc --pid ...
       └─ nomad exec2-shim <args>      ← already running as uid 80000
            └─ user command             ← CapAmb: 0000000000000000 always
```

**The kernel enforces a hard rule:** any call to `setresuid()` with a new non-zero real UID clears the entire ambient capability set. Because unshare --setuid calls setresuid() internally as part of its own uid transition, the ambient set was always wiped before the shim even started. There was no way to recover it.

_**Root cause**_

Kernel rule from [capabilities(7)](https://man7.org/linux/man-pages/man7/capabilities.7.html): A call to setresuid(nonzero) unconditionally clears the entire ambient set. The only way to preserve permitted capabilities across that transition is `prctl(PR_SET_KEEPCAPS, 1)` set before the uid change  which `unshare` does not do.

_**The Fix — Move the uid/gid Transition Into the Shim**_

`unshare --setuid` and `--setgid` are removed. The shim now receives the target uid, gid, and requested capability names as argv slots, and performs the uid/gid transition itself using the sequence prescribed by `capabilities(7)`:

```
Nomad agent (root)
  └─ unshare --ipc --pid --mount-proc --fork   ← no --setuid/--setgid
       └─ nomad exec2-shim ... uid gid caps ... ← still root at entry
            │  1. prctl(PR_SET_KEEPCAPS, 1)     ← preserve permitted across setresuid
            │  2. setresgid(gid)
            │  3. setresuid(uid)                ← now uid 80000, permitted still intact
            │  4. CAPSET(caps)                  ← restore in permitted+effective+inheritable
            │  5. prctl(PR_CAP_AMBIENT_RAISE)   ← raise as ambient
            │  6. prctl(PR_SET_KEEPCAPS, 0)     ← clean up
            └─ user command                     ← CapAmb: 0000000000000400 ✓
```

_**The dropPrivileges function — four-case matrix**_

uid | caps requested | Steps executed | Syscall count
-- | -- | -- | --
0 (root) | none | nothing | 0
0 (root) | [net_bind_service] | CAPSET + AMBIENT_RAISE (steps 4–5 only; permitted already full for root) | 2+n
80000 (non-root) | none | setresgid + setresuid (steps 2–3 only) | 2
80000 (non-root) | [net_bind_service] | Full sequence steps 1–6 (KEEPCAPS bridges the uid transition) | 4+2n

_**Argv protocol change**_

Slot | Before | After
-- | -- | --
argv[0..4] | binary, subcommand, defaults, stdout, stderr | same
argv[5] | unveil paths start here | <uid> (new)
argv[6] | -- sentinel | <gid> (new)
argv[7] | user command | cap,cap,... (new; "" if none)
argv[8+] | user args | unveil paths, --, user command, args (shifted +3)

_**New Configuration Fields**_

Field | Level | Type | Default | Description
-- | -- | -- | -- | --
allow_caps | Plugin config (operator) | list(string) | 13 Nomad-default caps | Operator allowlist. Tasks can only request capabilities present in this list.
alloc_caps | Task config (job author) | list(string) | [] (none) | Capabilities to grant as ambient to the task process. Must be a subset of allow_caps.

_**Capability names are case-insensitive and accept all four common formats:**_

```
net_bind_service        # lowercase, no prefix  ← canonical
NET_BIND_SERVICE        # uppercase, no prefix
cap_net_bind_service    # lowercase, with prefix
CAP_NET_BIND_SERVICE    # uppercase, with prefix  ← also accepted
```

_**Default allow_caps (13 caps — matches Nomad exec driver)**_
```
audit_write, chown, dac_override, fowner, fsetid, kill, mknod,
net_bind_service, setfcap, setgid, setpcap, setuid, sys_chroot
```

**_Testing_**
<details>
1) verify alloc_caps works end-to-end:

```
job "caps-test" {
  type = "batch"

  constraint {
    value     = "linux"
  }

  group "group" {
    reschedule {
      attempts  = 0
      unlimited = false
    }

    restart {
      attempts = 0
      mode     = "fail"
    }

    task "check-caps" {
      driver = "exec2"

      config {
        command = "sh"
        args    = ["-c", "grep CapAmb /proc/self/status"]

        # alloc_caps: grant CAP_NET_BIND_SERVICE as an ambient capability.
        # The kernel applies this after the uid/gid drop inside unshare,
        # so the unprivileged dynamic workload user inherits it.
        alloc_caps = ["net_bind_service"]

        # unveil /proc so the sandboxed task can read /proc/self/status
        unveil = ["r:/proc"]
      }

      resources {
        cpu    = 100
        memory = 32
      }
    }
  }
}
```
**Result:**
```
riteshharihar@podman-dev:/Users/riteshharihar/Desktop/Src/nomad-driver-exec2$ nomad job status caps-test
ID            = caps-test
Name          = caps-test
Submit Date   = 2026-07-31T15:11:40+05:30
Type          = batch
Priority      = 50
Datacenters   = *
Namespace     = default
Node Pool     = default
Status        = dead
Periodic      = false
Parameterized = false

Summary
Task Group  Queued  Starting  Running  Failed  Complete  Lost  Unknown
group       0       0         0        0       1         0     0

Allocations
ID        Node ID   Task Group  Version  Desired  Status    Created    Modified
31b04c09  8b821ad6  group       0        run      complete  6m21s ago  6m18s ago

riteshharihar@podman-dev:/Users/riteshharihar/Desktop/Src/nomad-driver-exec2$ nomad alloc logs 31b04c09
CapAmb: 0000000000000400
```

2) net_bind_service lets a task bind port 80

```
job "port80" {
  type = "service"

  constraint {
    value     = "linux"
  }

  group "group" {
    network {
      mode = "host"
      # Port 80 — requires CAP_NET_BIND_SERVICE for non-root processes
      port "http" { static = 80 }
    }

    task "server" {
      driver = "exec2"

      config {
        command = "python3"
        args    = ["-m", "http.server", "80", "--directory", "${NOMAD_TASK_DIR}"]

        # Without this, python3 (running as nomad-XXXXX, non-root) would fail:
        # OSError: [Errno 13] Permission denied (binding port 80)
        #
        # With this, CAP_NET_BIND_SERVICE is raised as an ambient capability,
        # so the kernel allows binding the privileged port.
        alloc_caps = ["net_bind_service"]
      }

      resources {
        cpu    = 200
        memory = 64
      }
    }
  }
}
```

**Result:** 
```
riteshharihar@podman-dev:/Users/riteshharihar/Desktop/Src/nomad-driver-exec2$ nomad alloc status f6fc691c
ID                  = f6fc691c-bc3c-aef1-ea4c-ed226ad39538
Eval ID             = 3b8fe0e7
Name                = port80.group[0]
Node ID             = ffef0122
Node Name           = podman-dev
Job ID              = port80
Job Version         = 0
Client Status       = running
Client Description  = Tasks are running
Desired Status      = run
Desired Description = <none>
Created             = 43s ago
Modified            = 32s ago
Deployment ID       = 80505c44
Deployment Health   = healthy

Allocation Addresses (mode = "host"):
Label  Dynamic  Address
*http  yes      127.0.0.1:80

Task "server" is "running"
Task Resources:
CPU        Memory         Disk     Addresses
0/200 MHz  61 MiB/64 MiB  300 MiB  

Task Events:
Started At     = 2026-07-31T10:05:58Z
Finished At    = N/A
Total Restarts = 0
Last Restart   = N/A

Recent Events:
Time                       Type        Description
2026-07-31T15:35:58+05:30  Started     Task started by client
2026-07-31T15:35:58+05:30  Task Setup  Building Task Directory
2026-07-31T15:35:58+05:30  Received    Task received by client


riteshharihar@podman-dev:/Users/riteshharihar/Desktop/Src/nomad-driver-exec2$ curl -s http://localhost:80/
<!DOCTYPE HTML>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Directory listing for /</title>
</head>
<body>
<h1>Directory listing for /</h1>
<hr>
<ul>
</ul>
<hr>
</body>
</html>
```

3) Capability not in allow_caps:

```
server {
  enabled          = true
  bootstrap_expect = 1
}

client {
  enabled = true
  options = {
    "fingerprint.denylist" = "env_aws,env_gce,env_azure,env_digitalocean"
  }
}

plugin "nomad-driver-exec2" {
  config {
    unveil_defaults = true
    unveil_by_task  = true
    # Empty allowlist: operator disallows all capability grants.
    # Any task that specifies alloc_caps will be rejected at start time.
    allow_caps = []
  }
}
```

Result:
```
riteshharihar@podman-dev:/Users/riteshharihar/Desktop/Src/nomad-driver-exec2$ nomad alloc status 91f4baf2
ID                  = 91f4baf2-70e2-13f3-8818-33b7a6d4c5b2
Eval ID             = 5dc338dc
Name                = caps-test.group[0]
Node ID             = d99708db
Node Name           = podman-dev
Job ID              = caps-test
Job Version         = 0
Client Status       = failed
Client Description  = Failed tasks
Desired Status      = run
Desired Description = <none>
Created             = 1m13s ago
Modified            = 1m9s ago

Task "check-caps" is "dead"
Task Resources:
CPU      Memory  Disk     Addresses
100 MHz  32 MiB  300 MiB  

Task Events:
Started At     = N/A
Finished At    = 2026-07-31T09:58:57Z
Total Restarts = 0
Last Restart   = N/A

Recent Events:
Time                       Type            Description
2026-07-31T15:28:57+05:30  Not Restarting  Policy allows no restarts
2026-07-31T15:28:57+05:30  Driver Failure  rpc error: code = Unknown desc = task requested capability "net_bind_service" not in driver allow_caps
2026-07-31T15:28:57+05:30  Task Setup      Building Task Directory
2026-07-31T15:28:57+05:30  Received        Task received by client
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+FEATURES:
+
+* Added `allow_caps` (plugin config), `cap_add` and `cap_drop` (task config) to support Linux ambient capabilities for tasks. [[GH-96](https://github.com/hashicorp/nomad-driver-exec2/pull/96)]
+
 BREAKING CHANGES:
 
 * Task working directory is now explicitly set to `$NOMAD_TASK_DIR`. Jobs using relative args (e.g. `args = ["local/run.sh"]`) must switch to absolute paths (e.g. `command = "${NOMAD_TASK_DIR}/run.sh"`). [[GH-97](https://github.com/hashicorp/nomad-driver-exec2/pull/97)]
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ plugin "nomad-driver-exec2" {
     unveil_defaults = true
     unveil_paths    = []
     unveil_by_task  = false
+    allow_caps      = ["net_bind_service", "chown", "kill", ...]
   }
 }
 ```
@@ -178,6 +179,24 @@ plugin "nomad-driver-exec2" {
   - `unveil_by_task` - (default: `false`) - enable or disable job submitters to
   specify additional filesystem path access within task config
 
+  - `allow_caps` - (default: 13 Nomad-default capabilities) - an operator
+  allowlist of Linux capability names that tasks on this node are permitted to
+  request. Capability names are case-insensitive and accept all common formats
+  (`net_bind_service`, `NET_BIND_SERVICE`, `CAP_NET_BIND_SERVICE`). Set to `[]`
+  to disallow all capability grants on this node.
+
+  ```hcl
+  # allow tasks to request net_bind_service only
+  allow_caps = ["net_bind_service"]
+
+  # disallow all capability grants
+  allow_caps = []
+  ```
+
+  The default set is:
+  `audit_write`, `chown`, `dac_override`, `fowner`, `fsetid`, `kill`, `mknod`,
+  `net_bind_service`, `setfcap`, `setgid`, `setpcap`, `setuid`, `sys_chroot`.
+
 #### Task Configuration
 
 ##### config
@@ -193,6 +212,8 @@ config {
   args          = ["/etc/os-release"]
   unveil        = ["r:/etc/os-release"]
   oom_score_adj = 500
+  cap_add       = ["net_bind_service"]
+  cap_drop      = []
 }
 ```
 
@@ -208,6 +229,27 @@ config {
 
   - `oom_score_adj` - (optional) - The likelihood of the task being OOM killed,
   must be a positive integer. Defaults to `0`.
+
+  - `cap_add` - (optional) - A list of Linux capability names to add as ambient
+  capabilities for the task process. The effective capability set starts empty;
+  `cap_add` adds to it. Each name must be present in the operator's `allow_caps`
+  plugin config or the task is rejected at start time. Capability names are
+  case-insensitive and accept all common formats (`net_bind_service`,
+  `NET_BIND_SERVICE`, `CAP_NET_BIND_SERVICE`). Defaults to `[]`.
+
+  - `cap_drop` - (optional) - A list of Linux capability names to remove from
+  the set assembled by `cap_add`. Names are normalized the same way as `cap_add`.
+  Useful for dropping a specific capability when `cap_add` was set broadly.
+  Defaults to `[]`.
+
+  ```hcl
+  # grant CAP_NET_BIND_SERVICE so the task can bind port 80 without root
+  cap_add = ["net_bind_service"]
+
+  # grant two caps then selectively revoke one
+  cap_add  = ["net_bind_service", "chown"]
+  cap_drop = ["chown"]
+  ```
 
 ##### cpu
 
@@ -239,6 +281,7 @@ can be used as constraints when authoring jobs.
 ```text
 driver.exec2.unveil.defaults    = true
 driver.exec2.unveil.tasks       = true
+driver.exec2.caps.allowlist     = audit_write,chown,dac_override,fowner,fsetid,kill,mknod,net_bind_service,setfcap,setgid,setpcap,setuid,sys_chroot
 ```
 
 ### Install
--- a/pkg/capabilities/capabilities.go
+++ b/pkg/capabilities/capabilities.go
@@ -1,0 +1,68 @@
+// Copyright IBM Corp. 2024, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package capabilities
+
+import (
+	"fmt"
+	"runtime"
+	"strings"
+
+	"github.com/moby/sys/capability"
+)
+
+// capNames maps normalized capability names (lowercase, no "cap_" prefix) to
+// their kernel integer values. Built at init time from the moby/sys/capability
+// library's authoritative list — automatically covering every capability the
+// library knows about, with no manual maintenance required.
+//
+// On Linux, ListSupported queries the running kernel for the actual supported
+// set. On other platforms, ListKnown returns the static compiled-in list
+// (used when cross-compiling or running tests on non-Linux).
+var capNames = func() map[string]uintptr {
+	var list []capability.Cap
+	switch runtime.GOOS {
+	case "linux":
+		list, _ = capability.ListSupported()
+	default:
+		list = capability.ListKnown()
+	}
+	m := make(map[string]uintptr, len(list))
+	for _, c := range list {
+		m[c.String()] = uintptr(c)
+	}
+	return m
+}()
+
+// ParseCap resolves a capability name to its kernel integer value. Names are
+// normalized before lookup: case-insensitive and with or without the "cap_"
+// prefix, so "NET_BIND_SERVICE", "cap_net_bind_service", "CAP_NET_BIND_SERVICE",
+// and "net_bind_service" all resolve to the same value.
+func ParseCap(name string) (uintptr, error) {
+	v, ok := capNames[NormalizeCap(name)]
+	if !ok {
+		return 0, fmt.Errorf("unknown capability %q", name)
+	}
+	return v, nil
+}
+
+// Resolve resolves a slice of capability names to their kernel integer values.
+// Returns the first error encountered for an unknown name.
+func Resolve(names []string) ([]uintptr, error) {
+	caps := make([]uintptr, 0, len(names))
+	for _, name := range names {
+		v, err := ParseCap(name)
+		if err != nil {
+			return nil, err
+		}
+		caps = append(caps, v)
+	}
+	return caps, nil
+}
+
+// NormalizeCap lowercases a capability name and strips any "cap_" prefix,
+// accepting all four common formats: "net_bind_service", "NET_BIND_SERVICE",
+// "cap_net_bind_service", "CAP_NET_BIND_SERVICE".
+func NormalizeCap(name string) string {
+	return strings.TrimPrefix(strings.ToLower(strings.TrimSpace(name)), "cap_")
+}
--- a/pkg/capabilities/capabilities.go
+++ b/pkg/capabilities/capabilities.go
@@ -12,10 +12,7 @@ import (
 )
 
 // capNames maps normalized capability names (lowercase, no "cap_" prefix) to
-// their kernel integer values. Built at init time from the moby/sys/capability
-// library's authoritative list — automatically covering every capability the
-// library knows about, with no manual maintenance required.
-//
+// their kernel integer values.
 // On Linux, ListSupported queries the running kernel for the actual supported
 // set. On other platforms, ListKnown returns the static compiled-in list
 // (used when cross-compiling or running tests on non-Linux).
--- a/pkg/shim/z_shim_cmd.go
+++ b/pkg/shim/z_shim_cmd.go
@@ -10,11 +10,32 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"runtime"
 	"strconv"
+	"strings"
+	"syscall"
+	"unsafe"
 
+	"golang.org/x/sys/unix"
+
+	"github.com/hashicorp/nomad-driver-exec2/pkg/capabilities"
 	"github.com/hashicorp/nomad-driver-exec2/pkg/util"
 	"github.com/hashicorp/nomad/helper/subproc"
 )
+
+// capHdr and capData mirror the kernel's __user_cap_header_struct and
+// __user_cap_data_struct used by capset(2). Version 3 (0x20080522) supports
+// capabilities 0–63 via two 32-bit words.
+type capHdr struct {
+	version uint32
+	pid     int32
+}
+
+type capData struct {
+	effective   uint32
+	permitted   uint32
+	inheritable uint32
+}
 
 const (
 	// SubCommand is the first argument to the clone of the nomad agent process
@@ -31,7 +52,7 @@ const (
 	ExitBadLogging = 41
 )
 
-// init is the entrypoint for the 'nomad e2e-shim' invocation of nomad
+// init is the entrypoint for the 'nomad exec2-shim' invocation of nomad
 //
 // The argument format is as follows,
 //
@@ -40,8 +61,11 @@ const (
 // 2. true/false       <- include default unveil paths
 // 3. <stdout path>    <- path to named pipe for standard output
 // 4. <stderr path>    <- path to named pipe for standard error
-// 5. [mode:path, ...] <- list of additional unveil paths
-// 6. --               <- sentinel between following commands
+// 5. <uid>            <- numeric user id for the task process
+// 6. <gid>            <- numeric group id for the task process
+// 7. cap,cap,...      <- comma-separated ambient capability names (empty string if none)
+// 8. [mode:path, ...] <- list of additional unveil paths
+// 9. --               <- sentinel between following commands
 func init() {
 	subproc.Do(SubCommand, func() int {
 		// we need to ignore the stop signal (which is sent to the entire
@@ -50,23 +74,56 @@ func init() {
 		sigs := make(chan os.Signal, 1)
 		signal.Notify(sigs)
 		go func() {
-			<-sigs // do nothing; say alive
+			<-sigs // do nothing; stay alive
 		}()
 
-		if n := len(os.Args); n <= 4 {
-			subproc.Print("failed to invoke e2e-shim with sufficient args: %d", n)
+		if n := len(os.Args); n <= 7 {
+			subproc.Print("failed to invoke exec2-shim with sufficient args: %d", n)
 			return ExitWrongArgs
 		}
 
-		// get the unveil paths and the rest of the command(s) to run
-		// from our command arguments
-		args := os.Args[5:] // chop off 'nomad exec2-shim <defaults>'
 		defaults := os.Args[2] == "true"
 		outPipePath := os.Args[3]
 		errPipePath := os.Args[4]
+		uid, err := strconv.Atoi(os.Args[5])
+		if err != nil {
+			subproc.Print("invalid uid %q: %v", os.Args[5], err)
+			return ExitWrongArgs
+		}
+		gid, err := strconv.Atoi(os.Args[6])
+		if err != nil {
+			subproc.Print("invalid gid %q: %v", os.Args[6], err)
+			return ExitWrongArgs
+		}
+		capsArg := os.Args[7]
+
+		// get the unveil paths and the rest of the command(s) to run
+		// from our command arguments (after uid, gid, and caps args)
+		args := os.Args[8:]
 		paths, commands := split(args)
 		paths = append(paths, "w:"+outPipePath)
 		paths = append(paths, "w:"+errPipePath)
+
+		// resolve capability names to kernel integers before calling dropPrivileges;
+		// this is a pure string-to-integer mapping that requires no privileges
+		var caps []uintptr
+		if capsArg != "" {
+			capNames := strings.Split(capsArg, ",")
+			var capErr error
+			caps, capErr = capabilities.Resolve(capNames)
+			if capErr != nil {
+				subproc.Print("failed to resolve capabilities: %v", capErr)
+				return ExitWrongArgs
+			}
+		}
+
+		// drop privileges conditionally: for non-root tasks, transitions uid/gid;
+		// for tasks with caps, restores them in permitted/effective/inheritable and
+		// raises as ambient so they survive into the task process via exec().
+		if err := dropPrivileges(uid, gid, caps); err != nil {
+			subproc.Print("failed to drop privileges: %v", err)
+			return subproc.ExitFailure
+		}
 
 		stdout, stderr, err := util.OpenPipes(outPipePath, errPipePath)
 		if err != nil {
@@ -94,7 +151,10 @@ func init() {
 			return subproc.ExitNotRunnable
 		}
 
-		// invoke the task command with its args
+		// invoke the task command with its args;
+		// ambient capabilities are already raised in this process and will be
+		// inherited by the child across the exec() boundary automatically
+		
 		// the environment has already been set for us by the exec2 driver;
 		// NOMAD_WORK_DIR is set to work_dir if configured, otherwise NOMAD_TASK_DIR
 		cmd := exec.Command(cmdpath, commands[1:]...)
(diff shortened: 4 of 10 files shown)


pkg/shim/capabilities.go
```diff
@@ -0,0 +1,85 @@
...
+	HCLSpecLiteral = `["audit_write","chown","dac_override","fowner","fsetid","kill","mknod","net_bind_service","setfcap","setgid","setpcap","setuid","sys_chroot"]`
+)
+
+var extractLiteral = regexp.MustCompile(`[\w]+`)
+
+// NomadDefaultCaps returns the default set of Linux capabilities allowed by
+// the exec2 driver plugin. It is derived directly from HCLSpecLiteral.
+func NomadDefaultCaps() []string {
+	return extractLiteral.FindAllString(HCLSpecLiteral, -1)
+}
```
tgross: You're defining a constant string that you consume in another package, and then you take that string here and build a regex to extract it into a string slice, and then that function is only ever used once, in a test assertion. Plus the name of the constant is `HCLSpecLiteral` which means nothing here in the context in which it's being defined.

This should be defined over in `plugin/about.go` where it's used in production code. If you have a test assertion here about a default configured over there, then the test is in the wrong place or not testing anything useful.

README.md
```diff
@@ -178,6 +179,24 @@ plugin "nomad-driver-exec2" {
...
+
+  ```hcl
+  # allow tasks to request net_bind_service only
+  allow_caps = ["net_bind_service"]
+
+  # disallow all capability grants
+  allow_caps = []
+  ```
+
+  The default set matches the Nomad built-in `exec` driver:
```
tgross: Conceivably that could change, so I would not note that these match. Just give the list.

README.md
```diff
@@ -193,6 +212,7 @@ config {
   args          = ["/etc/os-release"]
   unveil        = ["r:/etc/os-release"]
   oom_score_adj = 500
+  alloc_caps    = ["net_bind_service"]
```
tgross: My favorite typo

```suggestion
  allow_caps    = ["net_bind_service"]
```

README.md
```diff
@@ -209,6 +229,18 @@ config {
   - `oom_score_adj` - (optional) - The likelihood of the task being OOM killed,
   must be a positive integer. Defaults to `0`.
 
+  - `alloc_caps` - (optional) - A list of Linux capability names to grant as
```
tgross:
```suggestion
  - `allow_caps` - (optional) - A list of Linux capability names to grant as
```

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## UNRELEASED
 
+FEATURES:
+
+* Added `allow_caps` (plugin config) and `alloc_caps` (task config) to support Linux ambient capabilities for tasks. [[GH-96](https://github.com/hashicorp/nomad-driver-exec2/pull/96)]
```
tgross:
```suggestion
* Added `allow_caps` (plugin config) and `allow_caps` (task config) to support Linux ambient capabilities for tasks. [[GH-96](https://github.com/hashicorp/nomad-driver-exec2/pull/96)]
```

pkg/shim/capabilities.go
```diff
@@ -0,0 +1,85 @@
...
+		m[c.String()] = uintptr(c)
+	}
+	return m
+}()
+
+// ParseCap resolves a capability name to its kernel integer value. Names are
+// normalized before lookup: case-insensitive and with or without the "cap_"
+// prefix, so "NET_BIND_SERVICE", "cap_net_bind_service", "CAP_NET_BIND_SERVICE",
+// and "net_bind_service" all resolve to the same value.
+func ParseCap(name string) (uintptr, error) {
```
tgross, line 42: Do these functions need to be exported?

pkg/shim/capabilities_test.go
```diff
@@ -0,0 +1,115 @@
...
+	})
+
+	t.Run("first unknown name returns error", func(t *testing.T) {
+		_, err := ResolveCaps([]string{"chown", "bad_cap"})
+		must.EqError(t, err, `unknown capability "bad_cap"`)
+	})
+}
+
+func TestNomadDefaultCaps(t *testing.T) {
+	// every entry returned by NomadDefaultCaps must be resolvable
```
tgross: You're not resolving them! You're parsing them! Maybe you could just do both in one test? You've already got the "valid names resolved" test case above

plugin/driver_test.go
```diff
@@ -149,15 +149,17 @@ func TestFunctional_cases(t *testing.T) {
...
 		// task config
-		user    string
-		command string
-		args    []string
-		unveil  []string
+		user      string
+		command   string
+		args      []string
+		unveil    []string
+		allocCaps []string
```
tgross: Wait, did you really mean to call this `allocCaps`? This is a task driver, so it doesn't set capabilities on an allocation. These would be task caps. But `allow_caps` is going to match the user expectation better here.

tgross (changes requested):
I've left some comments about some details. But before I get into the meat of this, can you do me a favor and review the `unshare` man page for `--keep-caps`:

> When the --user option is given, ensure that capabilities granted in the user namespace are preserved in the child process.

This seems like the obvious approach to explore here. Maybe it's the wrong approach, but you didn't even comment on it in your PR description, which suggests there's some missing research.

ritesh-harihar:
> I've left some comments about some details. But before I get into the meat of this, can you do me a favor and review the `unshare` man page for `--keep-caps`:
> 
> > When the --user option is given, ensure that capabilities granted in the user namespace are preserved in the child process.
> 
> This seems like the obvious approach to explore here. Maybe it's the wrong approach, but you didn't even comment on it in your PR description, which suggests there's some missing research.

I did explored different approaches before moving forward with the current approach, but missed those to add on the PR description.

1) `SysProcAttr.AmbientCaps` on `unshare` process ->  but this fails because unshare --setuid calls setresuid(nonzero) internally → kernel clears ambient set

2) `SysProcAttr.AmbientCaps` on the inner exec.Command inside the shim -> but this also fails. Shim starts as uid 80000 with an empty permitted set. CAPSET (needed before raising ambient) requires `CAP_SETPCAP` in the effective set which the shim no longer has. Fails with EPERM.

3) User namespaces (`--user unshare` flag) -> `--user --keep-caps` gives the cap in the wrong namespace( the child one)  but the host network that exec2 tasks use is owned by the initial namespace, so the cap check fails exactly the same way as if no cap was granted at all.

`CAP_NET_BIND_SERVICE` inside a child user namespace does not allow binding privileged ports on a network namespace owned by the initial user namespace. The host network namespace is owned by the initial user namespace  and a child user namespace's caps cannot reach it. 

Additionally `--user` has a hard dependency on `kernel.unprivileged_userns_clone`, which is disabled on RHEL and hardened systems.

These were the reasons why went with the current approach.

README.md
ritesh-harihar: I think that there's a misunderstanding here.

**alloc_caps** is at `task config (job author)` level. Capabilities to grant as ambient to the task process. Must be  a subset of allow_caps.

**allow_caps** is at `Plugin config (operator)` level.  Tasks can only request capabilities present in this list.

so if we consider this i think there should be `alloc_caps` only.

README.md
ritesh-harihar: Same here, it should be **alloc_caps**  only

README.md
tgross: From my note on the other instance of it: this is a task driver, so it doesn't set capabilities on an _allocation_, it sets capabilities on a _task_. An allocation can have multiple tasks, each with a completely different driver. 

Note that under our typical operator personas the plugin and and the task are configued by two different people, so the name conflict doesn't matter that much. 

But if you're looking to differentiate the names, I'd at least match what we've done on the other drivers that set Linux capabilities like `docker` and `exec`: [`cap_add`](https://developer.hashicorp.com/nomad/docs/job-declare/task-driver/docker#cap_add) and [`cap_drop`](https://developer.hashicorp.com/nomad/docs/job-declare/task-driver/docker#cap_drop)

tgross:
> which is disabled on RHEL 

Landlock is also disabled on RHEL, making that irrelevant. But for some reason I thought we always set user namespaces here and that doesn't seem to be the case, so nevermind on that.

I'll re-review in that light.

README.md
ritesh-harihar: I had given a thought about using `cap_add` but since in the issue it was mentioned `alloc_caps` and since `cap_add` implies "add to an existing set" and exec2  has no default set to add to, so used `alloc_caps`.

I would be happy to rename `alloc_caps → cap_add` and document that exec2 has no `cap_drop` because it starts from zero. `cap_drop` would be a no-op field.

pkg/shim/capabilities.go
ritesh-harihar: yes `ParseCap`, `NormalizeCap`, and `HCLSpecLiteral` must stay exported as plugin package calls them directly.

README.md
tgross: > since in the issue it was mentioned `alloc_caps`

The issue did not. But even if it did, we shouldn't be blindly accepting a user's suggestion configuration without designing it in the context of the rest of the product.

> exec2 has no default set to add to

The default is the empty set! `cap_add` adds to the emtpy set. `cap_drop` drops from the set (which is handy if you support `"all"` in `cap_add`, as we do with the `docker` driver)

So to design this consistently with the other drivers:
* Start with the empty set.
* `task.config.cap_add` allows the job author to add to the empty set.
* `task.config.cap_drop` allows the job author to drop from the set that `cap_add` filled.
* The resulting values of the set must be a subset of `plugin.config.allow_caps`
* `plugin.config.allow_caps` has a default of  `[audit_write, chown, dac_override, fowner, fsetid, kill, mknod, net_bind_service, setfcap, setgid, setpcap, setuid, sys_chroot]`

ritesh-harihar: Got it. Will do the changes to make it consistent with other drivers.

pkg/shim/z_shim_cmd.go
```diff
@@ -40,8 +59,11 @@ const (
 // 2. true/false       <- include default unveil paths
 // 3. <stdout path>    <- path to named pipe for standard output
 // 4. <stderr path>    <- path to named pipe for standard error
-// 5. [mode:path, ...] <- list of additional unveil paths
-// 6. --               <- sentinel between following commands
+// 5. <uid>            <- numeric user id for the task process
+// 6. <gid>            <- numeric group id for the task process
+// 7. cap,cap,...      <- comma-separated ambient capability names (empty string if none)
```
tgross, line 65: I don't think this works. How do you submit an empty string in a command line arg? These shim args are getting long. Maybe we want to start thinking about having flags here for clarity?

pkg/shim/z_shim_cmd.go
```diff
@@ -50,24 +72,57 @@ func init() {
...
+
+		// get the unveil paths and the rest of the command(s) to run
+		// from our command arguments (after uid, gid, and caps args)
+		args := os.Args[8:]
 		paths, commands := split(args)
 		paths = append(paths, "w:"+outPipePath)
 		paths = append(paths, "w:"+errPipePath)
 
+		// resolve capability names before the uid/gid drop so we still have
+		// root permitted capabilities; an empty capsArg means no caps requested
```
tgross: This comment is misleading because it suggests we need to be root to resolve the names, but this is just mapping a list of strings to a list of different strings that represent the caps and requires no privileges.

pkg/shim/capabilities.go
tgross: But they're _only_ called from the plugin package, which makes their existence here dubious.

pkg/shim/z_shim_cmd.go
```diff
@@ -116,3 +172,77 @@ func init() {
 		return code
 	})
 }
+
+// dropPrivileges conditionally drops uid/gid and raises the given capabilities
+// as ambient in the current process.
```
tgross: "raises the given capabilities as ambient" has the distinct smell of LLM-generated stringing together of jargon without understanding, taken only from the name of the syscall and not how people normally talk about this kind of thing.

And give the context:


```suggestion
// dropPrivileges conditionally drops uid/gid and adds the given capabilities to the ambient set of the shim process so they can be passed to the task process.
```

pkg/shim/z_shim_cmd.go
```diff
@@ -116,3 +172,77 @@ func init() {
...
+	needDrop := uid != 0      // uid/gid transition only needed for non-root tasks
+	needCaps := len(caps) > 0 // cap management only needed when caps are requested
+
+	// step 1: keep permitted caps across the uid transition; only needed when
+	// both a uid drop AND capability raising are required — if we stay root,
+	// permitted is already full; if no caps are requested, KEEPCAPS is irrelevant
+	if needDrop && needCaps {
+		if _, _, errno := syscall.RawSyscall(syscall.SYS_PRCTL, unix.PR_SET_KEEPCAPS, 1, 0); errno != 0 {
+			return fmt.Errorf("prctl PR_SET_KEEPCAPS=1: %w", errno)
+		}
```
tgross, line 202: A general problem with this approach using `prctl`

> Set the state of the **calling thread's** "keep capabilities" flag.

(emphasis mine)

The shim is written in Go. Knowing that, how many threads do we have, and are all the syscalls you're making here guaranteed to run on the same thread? What should you do about that?

pkg/shim/z_shim_cmd.go
```diff
@@ -116,3 +172,77 @@ func init() {
...
+		if _, _, errno := syscall.RawSyscall(syscall.SYS_SETRESGID, uintptr(gid), uintptr(gid), uintptr(gid)); errno != 0 {
+			return fmt.Errorf("setresgid %d: %w", gid, errno)
+		}
+		if _, _, errno := syscall.RawSyscall(syscall.SYS_SETRESUID, uintptr(uid), uintptr(uid), uintptr(uid)); errno != 0 {
+			return fmt.Errorf("setresuid %d: %w", uid, errno)
+		}
+	}
+
+	// steps 4+5: restore caps in permitted/effective/inheritable then raise as ambient;
+	// for root the permitted set is already full so CAPSET still works without KEEPCAPS
```
tgross: If I've changed to a non-root UID/GID by this point, don't I need to also do `PR_SET_NO_NEW_PRIVS` so that the task we fork can't re-escalate via setuid/setgid binaries and set caps?

Otherwise, we'd also need to restrict the bounding set here after this call.

pkg/shim/z_shim_cmd.go
```diff
@@ -116,3 +172,77 @@ func init() {
...
+			data[idx].permitted |= mask
+			data[idx].inheritable |= mask
+		}
+		if _, _, errno := syscall.RawSyscall(syscall.SYS_CAPSET,
+			uintptr(unsafe.Pointer(&hdr)),
+			uintptr(unsafe.Pointer(&data[0])),
+			0,
+		); errno != 0 {
+			return fmt.Errorf("capset: %w", errno)
+		}
```
tgross, line 243: My understanding is that we only need this block in the case where `if need caps && needDrop`?

tgross (changes requested):

pkg/shim/capabilities.go
ritesh-harihar: `ParseCap()` is also called in `resolvecaps()` in capabilities.go . What i can think of is whether to keep it here only or keep the logic in both places  which would be redundant i guess.

pkg/shim/z_shim_cmd.go
ritesh-harihar: `How do you submit an empty string in a command line arg? `

so it will look something like this,

> cmd := exec.Command("nomad", "exec2-shim", "true", "/out", "/err", "1000", "1000", "", "r:/tmp", "--", "env")
> // os.Args[7] in the shim == ""
> // capsArg != "" is false → no caps resolved

The empty string at index 7 is passed exactly as-is to execve. No shell ever sees it, so there is nothing to quote or escape. The shim receives it as a zero-length string and the `if capsArg != ""` guard correctly skips cap resolution.

> Maybe we want to start thinking about having flags here for clarity?

yeah maybe we can create a separate issue for this.

pkg/shim/z_shim_cmd.go
ritesh-harihar: Yeah there can be multiple OS threads and all these syscalls are per-thread so without pinning, the goroutine can migrate between them mid-sequence, silently corrupting capability state.

Added `runtime.LockOSThread()` at the top of dropPrivileges, which guarantees all syscalls run on the same OS thread. Didn't unlock it because the shim process exits immediately, it either errors out or calls cmd.Run() and then returns an exit code so the  OS thread is destroyed with the process.

pkg/shim/z_shim_cmd.go
ritesh-harihar: Added `PR_SET_NO_NEW_PRIVS` in step 4 before capset.

pkg/shim/z_shim_cmd.go
ritesh-harihar: > A capability can be added to the ambient set only if it is already present in both the permitted and the inheritable sets of the process.

When uid=0 and caps=[net_bind_service]:

Permitted set: full (root has everything) 
Inheritable set: empty 

The kernel initialises the inheritable set to empty for every process, including root. Root's full permitted set does not automatically populate inheritable. So before calling `PR_CAP_AMBIENT_RAISE`, we must explicitly put net_bind_service into inheritable via `capset` otherwise the ambient raise fails with EPERM.

This is why the `capset` block must be inside if needCaps, not if needCaps && needDrop.

pkg/shim/z_shim_cmd.go
tgross: As per our discussion in the other PR around a config file, that'd be a better approach given how many config knobs we've added here.

pkg/capabilities/capabilities.go
```diff
@@ -0,0 +1,85 @@
...
+		m[c.String()] = uintptr(c)
+	}
+	return m
+}()
+
+// ParseCap resolves a capability name to its kernel integer value. Names are
+// normalized before lookup: case-insensitive and with or without the "cap_"
+// prefix, so "NET_BIND_SERVICE", "cap_net_bind_service", "CAP_NET_BIND_SERVICE",
+// and "net_bind_service" all resolve to the same value.
+func ParseCap(name string) (uintptr, error) {
```
tgross: That's fine. This is the sort of thing where I'd normally pull it all out into its own package (ex. `pkg/capabilities`), so that I'm not exposing `shim` internals directly to the top-level plugin package.

plugin/driver_test.go
```diff
@@ -439,6 +442,76 @@ func TestFunctional_cases(t *testing.T) {
...
+			args:           []string{"-c", "grep CapAmb /proc/self/status"},
+			unveilDefaults: true,
+			unveilPaths:    []string{"r:/proc"},
+			allowCaps:      []string{"net_bind_service"},
+			// capAdd intentionally omitted — no caps requested
+			exp:      &drivers.ExitResult{ExitCode: 0},
+			stdoutRe: regexp.MustCompile(`CapAmb:\s+0000000000000000`),
+		},
+		// cap names in mixed casing with CAP_ prefix must be accepted when
+		// allow_caps holds the lowercase unprefixed equivalent.
```
tgross, line 502: This is another good example of why we probably want to pull out a capabilities package: this is really just testing the normalization function in the shim package

pkg/shim/z_shim_cmd.go
tgross: 👍

tgross (commented):
Looking pretty good. I think it'd be nice to see if we can pull out `pkg/capabilities` if possible. It'll mean less importing the shim into the plugin. Don't forget that we'll need to update the website docs as well once this is ready to ship

plugin/driver_test.go
ritesh-harihar: Created new pkg/capabilities

tgross: Looks like you created a merge conflict in the process. I'd fix that and then squash this all down into a single commit with a good commit message

pkg/capabilities/capabilities.go
```diff
@@ -0,0 +1,68 @@
...
+	"runtime"
+	"strings"
+
+	"github.com/moby/sys/capability"
+)
+
+// capNames maps normalized capability names (lowercase, no "cap_" prefix) to
+// their kernel integer values. Built at init time from the moby/sys/capability
+// library's authoritative list — automatically covering every capability the
+// library knows about, with no manual maintenance required.
```
tgross:
```suggestion
// capNames maps normalized capability names (lowercase, no "cap_" prefix) to
// their kernel integer values
```

"built at init time... `<emdash>`... with no manual maintenance required"? Yeah, no kidding it's called an import. Write your own docstrings _if needed_ instead of having the LLM pour slop onto it. What's extra frustrating here is that _it's not even true_; the list is code-generated and committed into the upstream package, not init-time.

tgross (changes requested):

tgross (approved):

PR #97, opened 2026-08-04
fix: tasks are started with a cwd in a veiled directory

### LINKED ISSUE #83: Exec2 tasks are started with a cwd in a veiled directory
The exec2 task is starting inside the 'private' 'alloc/' dir instead of the exec2 'alloc_mounts/' dir.

As root on nomad test machine with nomad and nomad-driver-exec2 installed via apt. The
exec2 driver is in /opt/nomad/data/plugins via symlink and /usr/bin/ as installed.

data_dir is /opt/nomad-test/data
opt, nomad-test are root:root 755
data is root:root 700
as per: https://developer.hashicorp.com/nomad/docs/deploy/production/requirements#hardening-nomadnomad.hcl:

There must be a config mentioning the plugin so
nomad.hcl:
```
plugin "nomad-driver-exec2" {}
```

Start the nomad agent:
```
nomad agent -dev-connect -plugin-dir=/opt/nomad/data/plugins/ -config=nomad.hcl -data-dir=/opt/nomad-test/data
```

The job is being used is
exec2test.nomad:
```
exec2test.nomad:
job "e2job" {
  reschedule { # keep the logs cleaner
    attempts  = 0
    unlimited = false
  }
  type = "service"
  group "e2g" {
    count = 1
    task "e2t" {
      restart { # keep the logs cleaner
        attempts = 0
      }
      driver = "exec2"
      config {
        command = "bash"
        args    = ["local/run.sh"]
      }
      template {
        destination = "local/run.sh"
	perms = "0755"
        data = <<EOD
#!/bin/bash
set -x
# Keep all output interleaved :)
exec 2>&1
whoami
env | grep NOMAD_TASK_DIR
env | grep NOMAD_ALLOC_DIR

# This is a veiled directory that we should not (and can't) cd to.
pwd

cd local
pwd

cd ($pwd)

# Traverse a veiled directory (which we don't even have linux permission to enter when traversing from root in a hardened system which is odd) and arrive at an unveiled directory:
cd ../secrets
pwd

# Hang around
sleep 1d
EOD
      }
    }
  }
}
```

Run the job (purge any existing runs and ensure hardened):
```
nomad job stop -purge e2job
chmod 700 /opt/nomad-test/data/
nomad job run exec2test.nomad
```
This succeeds and the nomad logs contain:
```
[INFO]  client.driver_mgr.nomad-driver-exec2: exec2 runner: driver=exec2 oom_score_adj=0 unveil_defaults=true unveil_paths=[rwxc:/opt/nomad-test/alloc_mounts/0a53001c-1d6b-58cf-7556-7d237410a1cd-e2t/local, rwxc:/opt/nomad-test/alloc_mounts/0a53001c-1d6b-58cf-7556-7d237410a1cd-e2t/alloc, rwxc:/opt/nomad-test/alloc_mounts/0a53001c-1d6b-58cf-7556-7d237410a1cd-e2t/secrets, rwxc:/opt/nomad-test/alloc_mounts/0a53001c-1d6b-58cf-7556-7d237410a1cd-e2t/tmp] @module=exec2 args=[local/run.sh] cmd=bash timestamp=2026-03-19T12:46:46.338Z
```
and the job logs have:
```
+ whoami
whoami: cannot find name for user ID 80785
+ env
+ grep NOMAD_TASK_DIR
NOMAD_TASK_DIR=/opt/nomad-test/alloc_mounts/0a53001c-1d6b-58cf-7556-7d237410a1cd-e2t/local
+ env
+ grep NOMAD_ALLOC_DIR
NOMAD_ALLOC_DIR=/opt/nomad-test/alloc_mounts/0a53001c-1d6b-58cf-7556-7d237410a1cd-e2t/alloc
+ pwd
/opt/nomad-test/data/alloc/0a53001c-1d6b-58cf-7556-7d237410a1cd/e2t
+ cd local
+ pwd
/opt/nomad-test/data/alloc/0a53001c-1d6b-58cf-7556-7d237410a1cd/e2t/local
++ pwd
+ cd /opt/nomad-test/data/alloc/0a53001c-1d6b-58cf-7556-7d237410a1cd/e2t/local
local/run.sh: line 15: cd: /opt/nomad-test/data/alloc/0a53001c-1d6b-58cf-7556-7d237410a1cd/e2t/local: Permission denied
+ cd ../secrets
+ pwd
/opt/nomad-test/data/alloc/0a53001c-1d6b-58cf-7556-7d237410a1cd/e2t/secrets
+ sleep 1d
```

I recognise these are bind mounts to inodes that *are* unveiled and as such there is no security failing.

However having the working directory be a veiled directory feels wrong.

This file: https://github.com/hashicorp/nomad-driver-exec2/blob/main/pkg/shim/shim.go#L357

Says: 
`cmd.Dir = e.env.TaskDir`
which is not the same as:
`cmd.Dir = filepath.Dir(e.env.Env["NOMAD_TASK_DIR"])`

Which is used at https://github.com/hashicorp/nomad-driver-exec2/blob/main/pkg/shim/shim.go#L265 to find the directory to make the relative tmp/ dir

https://developer.hashicorp.com/nomad/docs/reference/runtime-environment-settings#task-directories
Says:
"The allocation working directory contains a task working directory for each task".
Now I understood this to be the linux "current working directory" (and for the other drivers this is the case).
Then it says: "relative to the task working directory" ... "alloc/" "local/" "secrets/"
Then it says that the path of "alloc/" is in NOMAD_ALLOC_DIR which finally pins us down to the exec2 task's value which we see in the output above.
Working back from NOMAD_ALLOC_DIR suggests the "task working directory" should be `filepath.Dir(e.env.Env["NOMAD_TASK_DIR"])` and that it would be both consistent with the docs and less surprising.

I do note that:
https://github.com/hashicorp/nomad/issues/8919 says
"The NOMAD_TASK_DIR is the local/ directory provided to each task, whereas the "task directory" is the directory that Nomad uses to set up the task, and may not be the actual filesystem of the task, "

If I patch the driver the output is:
```
+ whoami
whoami: cannot find name for user ID 89368
+ env
+ grep NOMAD_TASK_DIR
NOMAD_TASK_DIR=/opt/nomad-test/alloc_mounts/05762172-b640-dd44-0dbb-5d7e8678e6ed-e2t/local
+ env
+ grep NOMAD_ALLOC_DIR
NOMAD_ALLOC_DIR=/opt/nomad-test/alloc_mounts/05762172-b640-dd44-0dbb-5d7e8678e6ed-e2t/alloc
+ pwd
/opt/nomad-test/alloc_mounts/05762172-b640-dd44-0dbb-5d7e8678e6ed-e2t
+ cd local
+ pwd
/opt/nomad-test/alloc_mounts/05762172-b640-dd44-0dbb-5d7e8678e6ed-e2t/local
++ pwd
+ cd /opt/nomad-test/alloc_mounts/05762172-b640-dd44-0dbb-5d7e8678e6ed-e2t/local
+ cd ../secrets
+ pwd
/opt/nomad-test/alloc_mounts/05762172-b640-dd44-0dbb-5d7e8678e6ed-e2t/secrets
+ sleep 1d
```
and `cd $(pwd) ` is no longer an error :)

I'd be happy to submit the tiny PR if this is deemed correct.

ritesh-harihar:
Fixes: https://github.com/hashicorp/nomad-driver-exec2/issues/83

**Summary**

A task started by exec2 ran with its working directory set to the Nomad client's private internal alloc directory (`data/alloc/<id>/<task>`) rather than the unveiled bind-mount path (`$NOMAD_TASK_DIR`). Under Linux Landlock LSM, that private path is not in the unveil list, so any operation that reconstructs the CWD as a path string — `pwd`, `cd $(pwd)`, relative file writes — either produced the wrong path or failed with Permission denied on hardened systems.

The fix explicitly sets `cmd.Dir` on the inner task process to `$NOMAD_TASK_DIR`, matching the behaviour of Nomad's built-in raw_exec and exec drivers. An optional `work_dir` task config field is also added for consistency with those drivers.

**Root Cause**

**Two-process chain**
The exec2 driver runs tasks through a two-process chain. Both processes need their working directory set explicitly:

```
Nomad agent (root)
  └── nsenter → unshare → exec2-shim          ← outer process (shim.go / prepare())
                                   └── task binary   ← inner process (z_shim_cmd.go)
```

**What was missing**

`prepare()` in shim.go set` cmd.Dir = e.env.TaskDir `(the private path) for the outer shim process. `z_shim_cmd.go` built the inner task exec.Command with no `cmd.Dir `at all — so the task inherited whatever CWD the shim had after Landlock lockdown.

After lockdown() applied Landlock restrictions, the kernel could no longer traverse ancestor directories of the inherited CWD to reconstruct the path string, because those ancestors (data/alloc/…) were not in the unveil list. Only the bind-mount paths (`$NOMAD_TASK_DIR, $NOMAD_ALLOC_DIR…`) were unveiled.

**Additional Improvement — work_dir task config field**

As per discussion in the issue, exec2 should be consistent with Nomad's built-in drivers. All of `raw_exec`, `exec`, and `java` expose a work_dir task config field that overrides the default CWD. This PR adds the same.

```
task "example" {
  driver = "exec2"
  config {
    command  = "/usr/bin/my-service"
    work_dir = "/opt/my-service/data"   # optional; must be absolute
  }
}
```

Behaviour | Default (no work_dir) | With work_dir
-- | -- | --
Task CWD | $NOMAD_TASK_DIR | The specified absolute path
NOMAD_WORK_DIR env var | Set to $NOMAD_TASK_DIR | Set to the specified path
Landlock unveil | No extra unveil needed | Path auto-unveiled with rwxc regardless of unveil_defaults
Relative path | n/a | Rejected at task start: "work_dir must be an absolute path"

<br class="Apple-interchange-newline">

**Testing**
<details>

1) 
```
job "e2job" {
  reschedule {
    attempts  = 0
    unlimited = false
  }
  type = "service"
  group "e2g" {
    count = 1
    task "e2t" {
      restart {
        attempts = 0
      }
      driver = "exec2"
      config {
        command = "${NOMAD_TASK_DIR}/run.sh"
      }
      template {
        destination = "local/run.sh"
        perms       = "0755"
        data        = <<EOD
#!/bin/bash
set -x
exec 2>&1
whoami
env | grep NOMAD_TASK_DIR
env | grep NOMAD_ALLOC_DIR

# This is what pwd reports — should be $NOMAD_TASK_DIR but is the veiled private path
pwd

cd local
pwd

# This is the smoking-gun: cd $(pwd) should be a no-op but fails under Landlock
# because $(pwd) expands to the *private* unveil path not the alloc_mounts path
cd $(pwd)

# Traverse via relative .. — this still works (relative open from an already-open fd)
cd ../secrets
pwd

sleep 1d
EOD
      }

      resources {
        cpu    = 100
        memory = 64
      }
    }
  }
}
```

Result Before:
```
riteshharihar@podman-dev:/Users/riteshharihar/Desktop/Src/nomad-driver-exec2$ nomad alloc logs f0d109d5
+ whoami
whoami: failed to get username: No such id: 84269
+ env
+ grep NOMAD_TASK_DIR
NOMAD_TASK_DIR=/tmp/NomadClient3733059173/f0d109d5-5b38-487d-4aa6-a917e5718f8f-e2t/local
+ env
+ grep NOMAD_ALLOC_DIR
NOMAD_ALLOC_DIR=/tmp/NomadClient3733059173/f0d109d5-5b38-487d-4aa6-a917e5718f8f-e2t/alloc
+ pwd
/tmp/NomadClient3733059173/f0d109d5-5b38-487d-4aa6-a917e5718f8f/e2t
+ cd local
+ pwd
/tmp/NomadClient3733059173/f0d109d5-5b38-487d-4aa6-a917e5718f8f/e2t/local
++ pwd
+ cd /tmp/NomadClient3733059173/f0d109d5-5b38-487d-4aa6-a917e5718f8f/e2t/local
+ cd ../secrets
+ pwd
/tmp/NomadClient3733059173/f0d109d5-5b38-487d-4aa6-a917e5718f8f/e2t/secrets
+ sleep 1d
```

Result After Fix:
```
riteshharihar@podman-dev:/Users/riteshharihar/Desktop/Src/nomad-driver-exec2$ nomad alloc logs f5615d42
+ whoami
whoami: failed to get username: No such id: 88392
+ env
+ grep NOMAD_TASK_DIR
NOMAD_TASK_DIR=/tmp/NomadClient1676072496/f5615d42-1661-6e69-2782-229fd587e676-e2t/local
+ env
+ grep NOMAD_ALLOC_DIR
NOMAD_ALLOC_DIR=/tmp/NomadClient1676072496/f5615d42-1661-6e69-2782-229fd587e676-e2t/alloc
+ pwd
/tmp/NomadClient1676072496/f5615d42-1661-6e69-2782-229fd587e676-e2t/local
+ cd local
/tmp/NomadClient1676072496/f5615d42-1661-6e69-2782-229fd587e676-e2t/local/run.sh: line 11: cd: local: No such file or directory
+ pwd
/tmp/NomadClient1676072496/f5615d42-1661-6e69-2782-229fd587e676-e2t/local
++ pwd
+ cd /tmp/NomadClient1676072496/f5615d42-1661-6e69-2782-229fd587e676-e2t/local
+ cd ../secrets
+ pwd
/tmp/NomadClient1676072496/f5615d42-1661-6e69-2782-229fd587e676-e2t/secrets
+ sleep 1d
```

2)  relative file write succeeds
```
job "relwrite" {
  type = "batch"

  group "group" {
    reschedule {
      attempts  = 0
      unlimited = false
    }

    restart {
      attempts = 0
      mode     = "fail"
    }

    task "task" {
      driver = "exec2"

      config {
        command = "sh"
        args = [
          "-c",
          "echo hello > output.txt && cat output.txt"
        ]
      }

      resources {
        cpu    = 100
        memory = 32
      }
    }
  }
}
```

Result Berfore:
`/usr/bin/sh: 1: cannot create output.txt: Permission denied`

Result After Fix:
```
riteshharihar@podman-dev:/Users/riteshharihar/Desktop/Src/nomad-driver-exec2$ nomad alloc logs 14f62cff
hello
```

3) work_dir overrides the default CWD
```
job "work-dir" {
  type = "batch"

  group "group" {
    reschedule {
      attempts  = 0
      unlimited = false
    }

    restart {
      attempts = 0
      mode     = "fail"
    }

    task "task" {
      driver = "exec2"

      config {
        command = "sh"
        args = [
          "-c",
          "test \"$(pwd)\" = \"$NOMAD_ALLOC_DIR\" && echo WORKDIR_OK"
        ]
        work_dir = "${NOMAD_ALLOC_DIR}"
      }

      resources {
        cpu    = 100
        memory = 32
      }
    }
  }
}
```

Result:
```
riteshharihar@podman-dev:/Users/riteshharihar/Desktop/Src/nomad-driver-exec2$ nomad alloc logs e2016d35
WORKDIR_OK
```

4) work_dir rejects a relative path
```
job "work-dir-bad" {
  type = "batch"

  group "group" {
    task "task" {
      driver = "exec2"

      config {
        command = "pwd"
        work_dir = "relative/path"
      }

      resources {
        cpu    = 100
        memory = 32
      }
    }
  }
}
```

Result:
```
riteshharihar@podman-dev:/Users/riteshharihar/Desktop/Src/nomad-driver-exec2$ nomad alloc status 86562092
ID                   = 86562092-e86d-3418-5826-5e6fbfca72f5
Eval ID              = 81de2ed5
Name                 = work-dir-bad.group[0]
Node ID              = 616ee344
Node Name            = podman-dev
Job ID               = work-dir-bad
Job Version          = 0
Client Status        = failed
Client Description   = Failed tasks
Desired Status       = stop
Desired Description  = alloc was rescheduled because it failed
Created              = 7s ago
Modified             = 3s ago
Replacement Alloc ID = d831cd60

Task "task" is "dead"
Task Resources:
CPU      Memory  Disk     Addresses
100 MHz  32 MiB  300 MiB  

Task Events:
Started At     = N/A
Finished At    = 2026-08-04T06:52:44Z
Total Restarts = 0
Last Restart   = N/A

Recent Events:
Time                       Type            Description
2026-08-04T12:22:44+05:30  Not Restarting  Error was unrecoverable
2026-08-04T12:22:44+05:30  Driver Failure  rpc error: code = Unknown desc = work_dir must be an absolute path: "relative/path"
2026-08-04T12:22:44+05:30  Task Setup      Building Task Directory
2026-08-04T12:22:44+05:30  Received        Task received by client
```

</details>

### DIFF
--- a/pkg/shim/shim.go
+++ b/pkg/shim/shim.go
@@ -33,6 +33,7 @@ type Options struct {
 	UnveilPaths    []string
 	UnveilDefaults bool
 	OOMScoreAdj    int
+	WorkDir        string // working directory for the task; defaults to TaskDir
 }
 
 // Environment represents runtime configuration.
@@ -42,6 +43,7 @@ type Environment struct {
 	ErrPipe      string            // io pipe path for stderr
 	Env          map[string]string // environment variables
 	TaskDir      string            // task directory
+	WorkDir      string            // working directory for the task; defaults to TaskDir
 	Cgroup       string            // task cgroup path
 	Net          string            // allocation network namespace path
 	Memory       uint64            // memory in megabytes
@@ -293,7 +295,7 @@ func (e *exe) writeCG(file, content string) error {
 	return f.Close()
 }
 
-func flatten(user, home string, env map[string]string) []string {
+func flatten(user, home, workDir string, env map[string]string) []string {
 	result := make([]string, 0, len(env))
 
 	// override and remove some variables
@@ -313,6 +315,13 @@ func flatten(user, home string, env map[string]string) []string {
 	parent := filepath.Dir(env["NOMAD_TASK_DIR"])
 	tmp := filepath.Join(parent, "tmp")
 	env["TMPDIR"] = tmp
+
+	// set the working directory; defaults to NOMAD_TASK_DIR when not overridden
+	if workDir != "" {
+		env["NOMAD_WORK_DIR"] = workDir
+	} else {
+		env["NOMAD_WORK_DIR"] = env["NOMAD_TASK_DIR"]
+	}
 
 	// copy environment variables into list form
 	for k, v := range env {
@@ -425,8 +434,13 @@ func (e *exe) prepare(ctx context.Context, home string, fd, uid, gid int) (*exec
 	cmd.Stderr = errfd
 	e.errfd = errfd
 
-	cmd.Env = flatten(e.env.User, home, e.env.Env)
-	cmd.Dir = e.env.TaskDir
+	cmd.Env = flatten(e.env.User, home, e.env.WorkDir, e.env.Env)
+	// use work_dir when set, otherwise fall back to the task directory
+	if e.env.WorkDir != "" {
+		cmd.Dir = e.env.WorkDir
+	} else {
+		cmd.Dir = e.env.TaskDir
+	}
 	cmd.SysProcAttr = &syscall.SysProcAttr{
 		UseCgroupFD: true, // clone directly into cgroup
 		CgroupFD:    fd,   // cgroup file descriptor
--- a/pkg/shim/z_shim_cmd.go
+++ b/pkg/shim/z_shim_cmd.go
@@ -95,8 +95,10 @@ func init() {
 		}
 
 		// invoke the task command with its args
-		// the environment has already been set for us by the exec2 driver
+		// the environment has already been set for us by the exec2 driver;
+		// NOMAD_WORK_DIR is set to work_dir if configured, otherwise NOMAD_TASK_DIR
 		cmd := exec.Command(cmdpath, commands[1:]...)
+		cmd.Dir = os.Getenv("NOMAD_WORK_DIR")
 		cmd.Stdout = stdout
 		cmd.Stderr = stderr
 
--- a/plugin/driver.go
+++ b/plugin/driver.go
@@ -11,7 +11,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"slices"
-	"strings"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
@@ -561,11 +560,14 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
 	// is required. work_dir outside the alloc directory expands the filesystem
 	// surface, so it uses the same unveil_by_task gate as task-specified
 	// unveil paths.
+	//
+	// childEscapesParentDir uses os.OpenRoot so the kernel enforces the
+	// boundary — symlinks pointing outside the alloc root cannot bypass this.
 	if taskConfig.WorkDir != "" {
 		// alloc root is the grandparent of NOMAD_TASK_DIR:
 		// NOMAD_TASK_DIR = <alloc>/<task>/local  →  alloc root = <alloc>
 		allocRoot := filepath.Dir(filepath.Dir(driverTaskConfig.Env["NOMAD_TASK_DIR"]))
-		if !withinDir(allocRoot, taskConfig.WorkDir) {
+		if err := childEscapesParentDir(allocRoot, taskConfig.WorkDir); err != nil {
 			if !p.config.UnveilByTask {
 				return nil, fmt.Errorf("task set work_dir outside sandbox but driver config does not allow this")
 			}
@@ -592,10 +594,27 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
 	}, nil
 }
 
-// withinDir reports whether path is equal to dir or nested inside it.
-// It uses filepath.Rel so separator arithmetic is handled by the standard
-// library — a relative result starting with ".." means path is outside dir.
-func withinDir(dir, path string) bool {
-	rel, err := filepath.Rel(dir, path)
-	return err == nil && !strings.HasPrefix(rel, "..")
+// childEscapesParentDir reports whether child escapes parent by returning a
+// non-nil error. Uses os.OpenRoot so that the kernel enforces the sandbox
+// boundary, a symlink inside parent that points outside it cannot bypass
+// this check. This mirrors the escapingfs.ChildEscapesParentDir helper
+// introduced in Nomad v2.0.5
+func childEscapesParentDir(parent, child string) error {
+	if filepath.IsAbs(child) {
+		var err error
+		child, err = filepath.Rel(parent, child)
+		if err != nil {
+			return err
+		}
+	}
+	root, err := os.OpenRoot(parent)
+	if err != nil {
+		return err
+	}
+	defer root.Close()
+	_, err = root.Stat(child)
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	return nil
 }
--- a/plugin/driver.go
+++ b/plugin/driver.go
@@ -528,9 +528,13 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
 		return nil, fmt.Errorf("failed to decode driver task config: %w", err)
 	}
 
-	// work_dir must be an absolute path when set
+	// if work_dir is set, resolve a relative path against the parent of
+	// NOMAD_TASK_DIR (the task working directory: <alloc>/<task-name>/) so
+	// that job authors can write portable relative paths like "local/subdir"
+	// without needing to know the runtime absolute allocation path.
 	if taskConfig.WorkDir != "" && !filepath.IsAbs(taskConfig.WorkDir) {
-		return nil, fmt.Errorf("work_dir must be an absolute path: %q", taskConfig.WorkDir)
+		taskParent := filepath.Dir(driverTaskConfig.Env["NOMAD_TASK_DIR"])
+		taskConfig.WorkDir = filepath.Join(taskParent, taskConfig.WorkDir)
 	}
 
 	// combine paths to unveil from plugin config, task config (if enabled),
@@ -551,7 +555,12 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
 	// if work_dir is set, it must be accessible under Landlock — the task
 	// cannot chdir into a veiled directory. Auto-unveil it with rwxc so the
 	// task can read, write, execute, and create files there.
+	// work_dir_by_task must be enabled in the plugin config to allow job
+	// authors to set an arbitrary working directory.
 	if taskConfig.WorkDir != "" {
+		if !p.config.WorkDirByTask {
+			return nil, fmt.Errorf("task set work_dir but driver config does not allow this")
+		}
 		unveil = append(unveil, "rwxc:"+taskConfig.WorkDir)
 	}
 
@@ -573,3 +582,4 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
 		WorkDir:        taskConfig.WorkDir,
 	}, nil
 }
+
--- a/plugin/driver.go
+++ b/plugin/driver.go
@@ -253,6 +253,7 @@ func (p *Plugin) StartTask(config *drivers.TaskConfig) (*drivers.TaskHandle, *dr
 		ErrPipe:      errPipe,
 		Env:          config.Env,
 		TaskDir:      config.TaskDir().Dir,
+		WorkDir:      opts.WorkDir,
 		User:         config.User,
 		Cgroup:       cgroup,
 		Net:          netns(config),
@@ -326,6 +327,7 @@ func (p *Plugin) RecoverTask(handle *drivers.TaskHandle) error {
 		User:    handle.Config.User,
 		Cgroup:  cgroup,
 	}
+	// WorkDir is not needed for recovery (task is already running)
 
 	taskLogger := p.logger.With(
 		"alloc_id", taskState.TaskConfig.AllocID,
@@ -526,6 +528,11 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
 		return nil, fmt.Errorf("failed to decode driver task config: %w", err)
 	}
 
+	// work_dir must be an absolute path when set
+	if taskConfig.WorkDir != "" && !filepath.IsAbs(taskConfig.WorkDir) {
+		return nil, fmt.Errorf("work_dir must be an absolute path: %q", taskConfig.WorkDir)
+	}
+
 	// combine paths to unveil from plugin config, task config (if enabled),
 	// and some task/alloc directory default paths
 	unveil := slices.Clone(p.config.UnveilPaths)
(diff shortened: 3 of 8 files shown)


pkg/shim/z_shim_cmd.go
```diff
@@ -95,8 +95,10 @@ func init() {
 		}
 
 		// invoke the task command with its args
-		// the environment has already been set for us by the exec2 driver
+		// the environment has already been set for us by the exec2 driver;
+		// NOMAD_WORK_DIR is set to work_dir if configured, otherwise NOMAD_TASK_DIR
 		cmd := exec.Command(cmdpath, commands[1:]...)
+		cmd.Dir = os.Getenv("NOMAD_WORK_DIR")
```
ritesh-harihar, line 101: This is the main fix.

pkg/shim/shim.go
```diff
@@ -314,6 +316,13 @@ func flatten(user, home string, env map[string]string) []string {
 	tmp := filepath.Join(parent, "tmp")
 	env["TMPDIR"] = tmp
 
+	// set the working directory; defaults to NOMAD_TASK_DIR when not overridden
+	if workDir != "" {
+		env["NOMAD_WORK_DIR"] = workDir
+	} else {
+		env["NOMAD_WORK_DIR"] = env["NOMAD_TASK_DIR"]
+	}
```
tgross, line 324: This is fine but I'm beginning to think we have a large set of configuration options we're trying to pass thru the shim. Maybe we should think about generating a config file that the shim loads? That's how `runc` works.

We'd need to work out how we'd introduce that across task driver version upgrades, but maybe the shim for existing tasks doesn't care?

plugin/driver.go
```diff
@@ -541,6 +548,13 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
 		unveil = append(unveil, "rwxc:"+filepath.Join(parent, "tmp"))
 	}
 
+	// if work_dir is set, it must be accessible under Landlock — the task
+	// cannot chdir into a veiled directory. Auto-unveil it with rwxc so the
+	// task can read, write, execute, and create files there.
+	if taskConfig.WorkDir != "" {
+		unveil = append(unveil, "rwxc:"+taskConfig.WorkDir)
+	}
```
tgross, line 576: Does the cluster admin need to be able to control this via the plugin config? i.e. should the job author be able to bypass [`unveil_by_task`](https://developer.hashicorp.com/nomad/plugins/drivers/exec2#unveil_by_task)?

plugin/driver.go
```diff
@@ -526,6 +528,11 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
 		return nil, fmt.Errorf("failed to decode driver task config: %w", err)
 	}
 
+	// work_dir must be an absolute path when set
+	if taskConfig.WorkDir != "" && !filepath.IsAbs(taskConfig.WorkDir) {
+		return nil, fmt.Errorf("work_dir must be an absolute path: %q", taskConfig.WorkDir)
+	}
```
tgross: I understand that Landlock needs this to be an absolute path, but is there a reasonable knitting-together of a relative path with some base path we could use here? This could be relative to the task working directory (i.e. the parent of `NOMAD_TASK_DIR`, as described in the [filesystem docs](https://developer.hashicorp.com/nomad/docs/concepts/filesystem)), and I think that would let you avoid breaking backwards compatibility.

plugin/driver.go
ritesh-harihar: Added `work_dir_by_task` (default false) to the plugin config, mirroring the existing `unveil_by_task` pattern. When false, any task that sets `work_dir` is rejected at start time with a clear error. 

Keeping sandbox expansion fully under admin control `work_dir` auto-unveils its path, so added same kind of gate as unveil.

plugin/driver.go
ritesh-harihar: Done,  relative `work_dir` is now resolved against filepath.Dir(`NOMAD_TASK_DIR`)  rather than rejected.

pkg/shim/shim.go
ritesh-harihar: > Maybe we should think about generating a config file that the shim loads? We'd need to work out how we'd introduce that across task driver version upgrades

Yeah got this. Shall I keep the config-file refactor in this PR, or should I create a separate issue for it? I think it might require some additional effort and testing, so a follow-up might make more sense.

tgross: I would definitely have that as a follow-up. It's a major architectural change.

plugin/driver.go
tgross: I see where you're going with that, but now that's another thing to configure and it's all-or-nothing in the configuration. Knowing that the task working directory (i.e. `filepath.Dir(NOMAD_TASK_DIR)`)) is always unveiled, maybe this configuration is only needed if you have a `work_dir` path that's outside that sandbox? Seems redundant to restrict a task from setting `work_dir` if the location is already safe.

ritesh-harihar: Got the point.  Dropped `work_dir_by_task` entirely. If `work_dir` is inside the alloc directory it's already unveiled by defaults so no gate or extra config is needed. 

If it points outside the alloc (like `/tmp`) in that case it reuses the existing `unveil_by_task` gate since it's the same trust decision. Used [`filepath.Rel`]( https://pkg.go.dev/path/filepath#Rel) to keep the boundary check clean.

plugin/driver.go
```diff
@@ -556,5 +588,14 @@ func (p *Plugin) setOptions(driverTaskConfig *drivers.TaskConfig) (*shim.Options
...
 	}, nil
 }
+
+// withinDir reports whether path is equal to dir or nested inside it.
+// It uses filepath.Rel so separator arithmetic is handled by the standard
+// library — a relative result starting with ".." means path is outside dir.
+func withinDir(dir, path string) bool {
+	rel, err := filepath.Rel(dir, path)
+	return err == nil && !strings.HasPrefix(rel, "..")
+}
```
tgross: I'm pretty sure this is subject to symlink attacks where a second task in the same allocation can set up a symlink to outside the alloc root, bypassing the check. This is an unfortunately common problem for us in Nomad, enough that we have a helper function for it https://github.com/hashicorp/nomad/blob/v2.0.5/helper/escapingfs/escapes.go#L13-L49 which we're trying to migrate all similar checks to. I'd reuse that here.

tgross (approved):
LGTM

PR #98, opened 2026-08-07
fix: GOMAXPROCS auto-detection failure

### LINKED ISSUE #82: exec2 driver: GOMAXPROCS auto-detection fails in Go 1.25 due to cgroup namespacing
When using the exec2 driver with Go 1.25+ services, the Go runtime fails to auto-detect cgroup CPU quotas, defaulting instead to the total host core count. This occurs even when /sys/fs/cgroup and /proc are unveiled.

Observed Behavior
In Go 1.25, the runtime natively attempts to discover CPU limits via cgroup v2. However, inside an exec2 task, the process is placed in a new cgroup namespace.

/proc/self/cgroup reports the process is at the cgroup root (0::/).

Go runtime is not picking up the allocation's actual slice limit (nomad.slice/alloc-.../cpu.max), it defaults to the number of host cores causing overthreading.

Nomad v1.10.4
Go 1.25

ritesh-harihar:
Fixes: https://github.com/hashicorp/nomad-driver-exec2/issues/82

**_Summary_**

exec2 tasks saw `GOMAXPROCS` equal to the total host CPU count rather than their allocated CPU quota. This caused over-threading, excessive cgroup throttling, and degraded throughput for any Go binary scheduled with a fractional CPU share (e.g. cpu = 500 on a 32-core host). The driver now injects the correct value before the task process starts.

_**Root cause**_

The Go runtime auto-detects `GOMAXPROCS` at startup by reading two kernel interfaces. Both are silently blocked by Landlock inside an exec2 task. The Landlock allow-list even with `unveil_defaults = true`  does not include `/sys/fs/cgroup` or `/sys/devices/system/cpu`. So Go silently falls through to the host core count every time.
 
**_Fix_**
The fix injects `GOMAXPROCS` directly into the task environment in `StartTask()`, derived from the same bandwidth value already computed for writing to cpu.max. The formula uses integer ceiling division, so fractional shares round up to the next whole thread and reserved-core allocations (cores = N) produce an exact result because the per-core MHz cancels out. 

The injection is skipped if the operator has already set `GOMAXPROCS` explicitly in the job's env block. Because the value arrives as an environment variable, the Go runtime reads it via os.Getenv during runtime.init with no file I/O, and the kernel-enforced cgroup limit and the Go thread pool size are now always derived from the same source.

**_Why integer ceiling, not float64?_**
bandwidth is uint64. Converting to `float64` loses precision above 2^53. Integer ceiling division (n + d - 1) / d is exact for all representable values.

_**Data flow — how the value reaches the process**_

Screenshot 2026-08-07 at 12 48 38 PM

_**Job spec examples**_

**Scenario A — fractional share (cpu = N MHz)**
```
resources {
  cpu    = 500    # 500 MHz on a 3 GHz host → bandwidth=16666 → GOMAXPROCS=1
  memory = 256
}
```
**Scenario B — whole core reservation (cores = N)**
```
resources {
  cores  = 2     # 2 reserved cores → bandwidth=200_000 → GOMAXPROCS=2 (exact)
  memory = 512
}
```
**Scenario C — verify inside the task**
```
config {
  command = "printenv"
  args    = ["GOMAXPROCS"]
}
```

**_Testing_**
<details>

1) 
```
job "gomaxprocs-test" {
  type = "batch"

  constraint {
    value     = "linux"
  }

  group "g" {
    reschedule {
      attempts  = 0
      unlimited = false
    }

    restart {
      attempts = 0
      mode     = "fail"
    }

    task "check" {
      driver = "exec2"

      config {
        command = "gomaxprocs-check"
      }

      resources {
        cpu    = 500
        memory = 64
      }
    }
  }
}
```

Result before:
```
riteshharihar@podman-dev:/tmp/gomaxprocs-test$ nomad job run /tmp/gomaxprocs-job.hcl
==> 2026-08-06T14:14:26+05:30: Monitoring evaluation "0ef9ff6c"
    2026-08-06T14:14:26+05:30: Evaluation triggered by job "gomaxprocs-test"
    2026-08-06T14:14:27+05:30: Allocation "4233757b" created: node "2dd2ca81", group "g"
    2026-08-06T14:14:27+05:30: Evaluation status changed: "pending" -> "complete"
==> 2026-08-06T14:14:27+05:30: Evaluation "0ef9ff6c" finished with status "complete"
riteshharihar@podman-dev:/tmp/gomaxprocs-test$ sleep 5 && nomad logs -job gomaxprocs-test check
GOMAXPROCS=14 NumCPU=14
```

Result After:
```
riteshharihar@podman-dev:/tmp/gomaxprocs-test$ nomad job run /tmp/gomaxprocs-job.hcl
==> 2026-08-06T14:18:05+05:30: Monitoring evaluation "f487422e"
    2026-08-06T14:18:05+05:30: Evaluation triggered by job "gomaxprocs-test"
    2026-08-06T14:18:06+05:30: Allocation "b6dce476" created: node "876a183c", group "g"
    2026-08-06T14:18:06+05:30: Evaluation status changed: "pending" -> "complete"
==> 2026-08-06T14:18:06+05:30: Evaluation "f487422e" finished with status "complete"
riteshharihar@podman-dev:/tmp/gomaxprocs-test$ sleep 5 && nomad logs -job gomaxprocs-test check
GOMAXPROCS=1 NumCPU=14
```

2) 
```
job "gomaxprocs-cores" {
  type = "batch"

  constraint {
    value     = "linux"
  }

  group "g" {
    reschedule {
      attempts  = 0
      unlimited = false
    }

    restart {
      attempts = 0
      mode     = "fail"
    }

    task "check" {
      driver = "exec2"

      config {
        command = "gomaxprocs-check"
      }

      resources {
        cores  = 1
        memory = 64
      }
    }
  }
}
EOF
```

Before Result:
```
riteshharihar@podman-dev:/tmp/gomaxprocs-test$ nomad job run /tmp/gomaxprocs-cores-job.hcl
==> 2026-08-07T13:25:11+05:30: Monitoring evaluation "c0ad6310"
    2026-08-07T13:25:11+05:30: Evaluation triggered by job "gomaxprocs-cores"
    2026-08-07T13:25:12+05:30: Allocation "efa53783" created: node "9a97a483", group "g"
    2026-08-07T13:25:12+05:30: Evaluation status changed: "pending" -> "complete"
==> 2026-08-07T13:25:12+05:30: Evaluation "c0ad6310" finished with status "complete"
riteshharihar@podman-dev:/tmp/gomaxprocs-test$ nomad logs -job gomaxprocs-cores check
GOMAXPROCS=1 NumCPU=1
```

After Result: 
```
riteshharihar@podman-dev:/tmp/gomaxprocs-test$ nomad job run /tmp/gomaxprocs-cores-job.hcl
==> 2026-08-07T13:35:29+05:30: Monitoring evaluation "0f1f72f9"
    2026-08-07T13:35:29+05:30: Evaluation triggered by job "gomaxprocs-cores"
    2026-08-07T13:35:29+05:30: Allocation "5f0ccb66" created: node "0018293f", group "g"
    2026-08-07T13:35:30+05:30: Allocation "5f0ccb66" status changed: "pending" -> "running" (Tasks are running)
    2026-08-07T13:35:30+05:30: Evaluation status changed: "pending" -> "complete"
==> 2026-08-07T13:35:30+05:30: Evaluation "0f1f72f9" finished with status "complete"
riteshharihar@podman-dev:/tmp/gomaxprocs-test$ nomad logs -job gomaxprocs-cores check
GOMAXPROCS=1 NumCPU=1
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 BUG FIXES:
 
+* Fixed `GOMAXPROCS` to prevent Go workloads from being over-threaded against the host CPU capacity. [[GH-98](https://github.com/hashicorp/nomad-driver-exec2/pull/98)]
 * Error messages from the `unshare`/`nsenter` shim processes now appear in the allocation logs. [[GH-95](https://github.com/hashicorp/nomad-driver-exec2/pull/95)]
 
 ## 0.1.2 (May 12, 2026)
--- a/pkg/shim/sandbox.go
+++ b/pkg/shim/sandbox.go
@@ -52,6 +52,11 @@ func lockdown(defaults bool, elements []string) error {
 			landlock.Dir("/usr/bin", "rx"),
 			landlock.Dir("/usr/local/bin", "rx"),
 		)
+		// expose /proc read-only so runtimes (Go 1.25+, JVM, dotnet) can read
+		// /proc/self/cgroup and /proc/self/mountinfo to discover their cgroup
+		// CPU and memory limits. unshare --mount-proc creates 
+		// an isolated /proc scoped to the task's PID namespace.
+		paths = append(paths, landlock.Dir("/proc", "r"))
 	}
 
 	return landlock.New(paths...).Lock(landlock.Mandatory)
--- a/plugin/driver.go
+++ b/plugin/driver.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
@@ -228,6 +229,22 @@ func (p *Plugin) StartTask(config *drivers.TaskConfig) (*drivers.TaskHandle, *dr
 	// get our assigned cpuset cores
 	cpuset := config.Resources.LinuxResources.CpusetCpus
 	p.logger.Trace("resources", "memory", memory, "memory_max", memoryMax, "compute", bandwidth, "cpuset", cpuset)
+
+	// compute and inject GOMAXPROCS so Go workloads see the correct parallelism
+	// limit rather than defaulting to the total host CPU count. The Go runtime
+	// cannot reliably auto-detect this inside an exec2 task because Landlock
+	// restricts access to /sys/fs/cgroup and /sys/devices/system/cpu by default.
+	//
+	// bandwidth is the cgroup cpu.max quota in microseconds per 100_000 µs period.
+	// Ceiling integer division by the period gives the whole-core equivalent:
+	//   cores=N tasks:   bandwidth = N × 100_000 → result = N (exact)
+	//   cpu=M MHz tasks: bandwidth = M×100_000/per_core_MHz → result = ceil(fraction)
+	//
+	// Only inject if the operator has not already set it explicitly.
+	if _, alreadySet := config.Env["GOMAXPROCS"]; !alreadySet {
+		gomaxprocs := max(1, int((bandwidth+99_999)/100_000))
+		config.Env["GOMAXPROCS"] = strconv.Itoa(gomaxprocs)
+	}
 
 	// with cgroups v2 this is just the task cgroup
 	cgroup := config.Resources.LinuxResources.CpusetCgroupPath
--- a/plugin/driver.go
+++ b/plugin/driver.go
@@ -231,9 +231,13 @@ func (p *Plugin) StartTask(config *drivers.TaskConfig) (*drivers.TaskHandle, *dr
 	p.logger.Trace("resources", "memory", memory, "memory_max", memoryMax, "compute", bandwidth, "cpuset", cpuset)
 
 	// compute and inject GOMAXPROCS so Go workloads see the correct parallelism
-	// limit rather than defaulting to the total host CPU count. The Go runtime
-	// cannot reliably auto-detect this inside an exec2 task because Landlock
-	// restricts access to /sys/fs/cgroup and /sys/devices/system/cpu by default.
+	// limit rather than defaulting to the total host CPU count.
+	//
+	// Go 1.25+ can read cpu.max from the task's own cgroup scope (now always
+	// unveiled read-only via setOptions) and auto-detect correctly. However Go
+	// 1.24 and below never read cgroup at all — they fall back to
+	// sched_getaffinity() which returns all host cores for cpu=N MHz tasks.
+	// The env var injection covers all Go versions uniformly.
 	//
 	// bandwidth is the cgroup cpu.max quota in microseconds per 100_000 µs period.
 	// Ceiling integer division by the period gives the whole-core equivalent:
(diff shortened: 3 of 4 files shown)


tgross (changes requested):
Rather than fixing this only for Go, shouldn't we fix this by changing the Landlock config? That way you can use this for anything. Ex. Java applications might want to read their own cgroup to learn how much memory they have allocated.

ritesh-harihar:
> Rather than fixing this only for Go, shouldn't we fix this by changing the Landlock config? That way you can use this for anything. Ex. Java applications might want to read their own cgroup to learn how much memory they have allocated.

Adding `/sys/fs/cgroup` to the Landlock defaults would expose the entire cgroup hierarchy to every task. Landlock only grants access to a directory subtree, so every task would gain read access to every other task's memory.current, cpu.stat etc across all co-located allocations. 

I thought this shouldn't happen that's why didn't touched landlock defaults.

> Java applications might want to read their own cgroup to learn how much memory they have allocated.

If we want runtimes other than Go, to be able to read their own cgroup limits  then we can introduce a new `unveil_cgroup = true` plugin flag in agent.hcl. Instead of exposing the entire `/sys/fs/cgroup` tree, the driver would append only the task's own cgroup scope path `CpusetCgroupPath`, which is already available in the driver at task start time. This would be safe, scoped to the task's own data only, and language-agnostic. 

Let me know if this makes sense.

tgross:
> I thought this shouldn't happen that's why didn't touched landlock defaults.

Right, but can't we expose just the subtree for this PID?

ritesh-harihar:
> Right, but can't we expose just the subtree for this PID?

Yeah,  but it cannot be done inside Landlock's setup phase as it has no knowledge of the task's runtime `cgroup` path and don’t know the actual cgroup path for a particular task.

The task-specific `cgroup` path is only available at `StartTask(`) in the driver, which runs separately from the shim.

So `unveil_cgroup` will basically bridge this gap. The driver gets the `CpusetCgroupPath` at runtime and passes that specific path to the shim as an unveil path. This way, Landlock only exposes the cgroup directory for that task instead of giving access to the entire cgroup tree. 

If this feels right  i will start making the changes.

tgross:
Should we even bother making this a configuration knob? What's the harm in exposing it (read-only!) to the task by default?

_edit_ but otherwise, yeah, that seems like the right approach

ritesh-harihar:
> Should we even bother making this a configuration knob? What's the harm in exposing it (read-only!) to the task by default?
> 
> _edit_ but otherwise, yeah, that seems like the right approach

No, we don't need any config knobs here; this can be tied to `unveil_defaults`.

Exposing `/proc` read-only so runtimes (Go 1.25+, JVM) can read /`proc/self/cgroup` and `/proc/self/mountinfo` to discover their cgroup CPU and memory limits. unshare --mount-proc creates an isolated `/proc` scoped to the task's PID namespace. Without `/proc` unveiled, Landlock blocks both reads and Go never reaches cpu.max. 

Go 1.24 and below never read cgroup so env-var injection is the fix for this(already present), and it gives `GOMAXPROCS=1` for allocations where Go 1.25+'s native minimum is 2 . 

Since Go checks the env-var first and skips all cgroup reads if set, the injection always wins for Go, the cgroup unveil is purely for Java and other runtimes. 

If we want to drop the support for Go 1.24 we can remove this block of code?

> if _, alreadySet := config.Env["GOMAXPROCS"]; !alreadySet {
> 		gomaxprocs := max(1, int((bandwidth+99_999)/100_000))
> 		config.Env["GOMAXPROCS"] = strconv.Itoa(gomaxprocs)
> 	}

tgross:
> If we want to drop the support for Go 1.24 we can remove this block of code?

Go 1.24 is out of support itself. But I also have no idea why we'd be special-casing environment variables for Go in this driver at all. What makes Go special? Are there well-known environment variables that Haskell can use to set threadpools too? What about Luau? :grin: I think you get my point: we should be supplying a generic facility that well-behaved runtimes can use (cgroups), not special-casing environment variables for particular runtimes.

ritesh-harihar:
Agreed. We should provide generic cgroup facility for all runtimes, rather than runtime-specific handling. Removed the GOMAXPROCS env var injection.

tgross (approved):
LGTM

